PR #18, opened 2021-12-20
Media Composition

pleek91:
reactive implementation of matchMedia

### DIFF
--- a/src/media/media.ts
+++ b/src/media/media.ts
@@ -10,11 +10,12 @@ export function media(query: Ref<string> | string): Ref<boolean> {
     }
 
     mediaQuery.addEventListener('change', updateMatches)
-
+    
     if(isRef(query)) {
         unwatch = watch(query, () => {
             mediaQuery.removeEventListener('change', updateMatches)
             mediaQuery = window.matchMedia(unref(query))
+            mediaQuery.addEventListener('change', updateMatches)
         })
     }
 
--- a/src/media/media.ts
+++ b/src/media/media.ts
@@ -1,0 +1,32 @@
+import { getCurrentInstance, isRef, onUnmounted, ref, Ref, unref, watch } from "vue"
+
+export function media(query: Ref<string> | string): Ref<boolean> {
+    let mediaQuery = window.matchMedia(unref(query))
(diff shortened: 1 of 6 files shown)


src/media/media.ts
```diff
@@ -0,0 +1,32 @@
...
+    function updateMatches(event: MediaQueryListEvent) {
+        matches.value = event.matches
+    }
+
+    mediaQuery.addEventListener('change', updateMatches)
+
+    if(isRef(query)) {
+        unwatch = watch(query, () => {
+            mediaQuery.removeEventListener('change', updateMatches)
+            mediaQuery = window.matchMedia(unref(query))
```
stackoverfloweth, line 17: not sure if intentional, but it seems like 
 - you add an event listener on 'change' line 12
 - then if `query` is a ref you add a watcher
 - once that watch fires at least once you remove the event listener from 12
 - you do NOT re-add the event listener on 'change'

tests/media/media.spec.ts
```diff
@@ -0,0 +1,28 @@
...
+      });
+    });
+
+    it('returns a ref', async () => {
+        const match = media('(hover)')
+
+        expect(isRef(match)).toBe(true)
+    })
+
+})
```
stackoverfloweth, line 28: does it make sense to test that on unmount it will remove event listener? I know you have other tests with vue components for testing unmount, just not sure how difficult it would be to check for existence of your event listener

stackoverfloweth, line 28: does it make sense to test that when the passed in `query` changes, the event listener is updated? Basically testing the callback of the watcher?

stackoverfloweth (approved):

src/media/media.ts
pleek91: Great catch. Fixed

tests/media/media.spec.ts
pleek91: I was unable to figure out how to test this. I cannot spy on `unwatch` directly because its essentially private. Maybe you have some ideas here?

pleek91: Same problem as the watcher. Its private and I not able to actually test that it happens

PR #24, opened 2022-02-18
Composition Aliases and Subscription Quality of Life Improvements

pleek91:
# Description
Its a community standard for compositions to follow the `useThing` naming convention. Adding aliases for both `media` and `subscribe` that use this naming convention. `media` and `subscribe` are now deprecated in favor of `useMedia` and `useSubscription`

Included in this PR are a couple quality of life improvements around creating subscriptions. When using `useSubscription` the `args` argument is no longer required when the action has no required arguments. This means it's no longer necessary to pass in an empty array like `useSubscription(myAction, [])`. You can just use `useSubscription(myAction)`

Using `useSubscription` with classes or objects has been inconvenient because of the binding issue with actions. This PR introduces a new utility called `createActions` that helps make using classes with `useSubscription` more convenient.

## Example
Currently this is how we export services
```typescript
class MyService {
  private greeting = 'hello'

  public greet() {
    return this.greeting // this line is the issue. `this` context gets changed when the action is executed in a subscription.
  }
}

export const myService = new MyService()
```
And because of the context issue this is how we're using services with subscriptions

```typescript
import { myService } from '@/services/MyService'
import { useSubscription } from '@prefecthq/vue-compositions'

const subscription = useSubscription(myService.greet.bind(myService))
```
Having to bind the class to the method is annoying. Using the `createActions` utility we can bind the correct context ahead of time. Changing the `mySerivce` export to this
```typescript
import { createActions } from '@prefecthq/vue-compositions'`

class MyService {
  ...
}

export const myService = createActions(new MyService())
```
Means we can skip the binding at implementation
```typescript
import { myService } from '@/services/MyService'
import { useSubscription } from '@prefecthq/vue-compositions'

const subscription = subscribe(myService.greet)
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,5 +7,5 @@ npm i --save @prefecthq/vue-compositions
 ```
 
 ## Compositions
-- [media](https://github.com/PrefectHQ/vue-compositions/tree/main/src/media)
-- [subscribe](https://github.com/PrefectHQ/vue-compositions/tree/main/src/subscribe)
+- [useMedia](https://github.com/PrefectHQ/vue-compositions/tree/main/src/media)
+- [useSubscription](https://github.com/PrefectHQ/vue-compositions/tree/main/src/subscribe)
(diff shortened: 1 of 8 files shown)


znicholasbrown (approved):
This is great!

PR #39, opened 2022-02-28
Fix response undefined and vue watch warnings

pleek91:
# Description
When a subscription contains reactive arguments a couple of unexpected things can happen.

1. When arguments change the subscriptions channel changes. If a new channel has been created the subscriptions response would be undefined until the action resolves. 

2. When a new subscription to an existing channel is creating the response is undefined until the action is executed again.

3. If not all the arguments are reactive, vue would log a warning in the console. Ultimately the subscription was working as expected (we don't want vue to update the subscription if a non-reactive argument is changed anyway). But this causes confusion.

## Solution to undefined subscription responses to new channels
When a subscriptions channel is changed, if that channel is new it's response will be `undefined`. When that is the case, retain the previous channels response. That way the response smoothly transitions from channel to channel rather than reverting to `undefined` response in between.

## Solution to undefined subscription response existing channels
Storing the response on the channel and passing it to the new subscription when it is created. This means a subscription to an existing channel immediately receives the existing response. 

## Solution to vue watch warnings on arguments
Filter down the arguments to only those that are watchable. That way vue isn't asked to watch an argument it cannot that isn't expected to be watch anyway.

### DIFF
--- a/src/subscribe/subscribe.ts
+++ b/src/subscribe/subscribe.ts
@@ -2,6 +2,7 @@ import { getCurrentInstance, isReactive, isRef, onUnmounted, shallowReactive, un
 import Manager from './manager'
 import Subscription from './subscription'
 import { Action, ActionArguments, SubscribeArguments, SubscriptionOptions } from './types'
+import { watchableArgs } from './utilities'
 
 const defaultManager = new Manager()
 
@@ -25,9 +26,11 @@ export function subscribe<T extends Action>(
     (unref(args) as Parameters<T>).some(isRef) ||
     (unref(args) as Parameters<T>).some(isReactive)
   ) {
+    const argsToWatch = watchableArgs(args)
+
     unwatch = watch(
-      args,
-      (newArgs) => {
+      argsToWatch,
+      () => {
         if (!subscription.isSubscribed()) {
           unwatch!()
           return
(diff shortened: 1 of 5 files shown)


src/subscribe/subscribe.ts
```diff
@@ -25,16 +26,23 @@ export function subscribe<T extends Action>(
...
           unwatch!()
           return
         }
 
         subscription.unsubscribe()
-        Object.assign(subscription, manager.subscribe(action, newArgs, options))
+
+        const newSubscription = manager.subscribe(action, args, options)
+
+        newSubscription.response.value ??= subscription.response.value
```
znicholasbrown, line 43: ❤️

tests/subscribe/useSubscription.spec.ts
```diff
@@ -366,4 +394,43 @@ describe('subscribe', () => {
...
+    jest.runAllTimers()
+    jest.useRealTimers()
+
+    await timeout()
+
+    valueArg.value = 1
+
+    await timeout()
+
+    expect(subscription.response.value).toBe(originalValue)
```
znicholasbrown, line 417: This test breaks my brain but I assume it works because `timeout` is flushing at 0ms and even though the value of `valueArg` is changing the subscription channel it proves the original value is still available after the scheduler flush but before the resolution at 100ms?

pleek91: I simplified this in the most recent commit. The timeout on 411 wasn't necessary. But essentially use fake timers and run the timers to execute the action. Then revert back to real timers. With real timers the setTimeout in the asyn action doesn't execute until after the `expect` which means we can verify the response is correct.

PR #46, opened 2022-05-02
Stop interval if action errors

pleek91:
# Description
If an action throws an error we don't need to keep executing the action as the response will not change. 

Also modified how channels and subscriptions keep track of the public date (loading, executed, response, etc) so that the channel has a copy. That way when a subscription is created it can be created with the data the channel already has. Previously only response and executed were populated when a subscription was created.

### DIFF
--- a/src/subscribe/channel.ts
+++ b/src/subscribe/channel.ts
@@ -42,12 +42,13 @@ export default class Channel<T extends Action = Action> {
   private readonly action: T
   private readonly args: ActionArguments<T>
   private readonly subscriptions: Map<number, Subscription<T>> = new Map()
-  private response: ActionResponse<T> | undefined = undefined
   private timer: ReturnType<typeof setInterval> | null = null
   private lastExecution: number = 0
   private _loading: boolean = false
   private _executed: boolean = false
   private _errored: boolean = false
+  private _error: unknown = null
+  private _response: ActionResponse<T> | undefined = undefined
 
   public constructor(manager: Manager, action: T, args: ActionArguments<T>) {
     this.signature = ChannelSignatureManager.get(action, args)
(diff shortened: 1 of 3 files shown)


src/subscribe/channel.ts
```diff
@@ -135,12 +170,13 @@ export default class Channel<T extends Action = Action> {
...
 
     try {
-      this.setResponse(await this.action(...args))
+      this.response = await this.action(...args)
       this.errored = false
       this.error = null
     } catch (error) {
       this.errored = true
       this.error = error
+      this.clearTimer()
```
pleek91, line 179: This makes sure the action isn't executed again if the action resulted in an error.

src/subscribe/subscription.ts
```diff
@@ -22,11 +22,16 @@ export default class Subscription<T extends Action> {
...
     this.id = SubscriptionIdManager.get()
     this.channel = channel
     this.options = options
-    this.response.value = response
+
+    this.loading.value = channel.loading
+    this.response.value = channel.response
+    this.errored.value = channel.errored
+    this.error.value = channel.error
+    this.executed.value = channel.executed
```
pleek91, line 34: This change makes sure subscriptions are initialized with all the data from the channel and not just the response.

stackoverfloweth (approved):

madkinsz:
Not sure how this fits in here, but what if the server returns a rate limit error or some other intermittent failure?

pleek91:
> Not sure how this fits in here, but what if the server returns a rate limit error or some other intermittent failure?

@madkinsz that's a good question. In that case a page refresh would attempt the calls again. Also, where this is implemented we can check the error and make a decision if we want to do something special for a specific call like trying again because of a rate limit or something.

PR #57, opened 2022-05-24
Feature: new composition `useElementWidth

stackoverfloweth:
# useElementWidth
The `useElementWidth` composition abstracts away the logic for tracking the client width of an HTMLElement, even when the client width changes on resize thanks to `useResizeObserver`.

## Example
```typescript
import { useElementWidth } from '@prefecthq/vue-compositions'

const templateRef = ref<HTMLElement | undefined>()
const widthInPx = useElementWidth(templateRef)
```

## Arguments
| element | `HTMLElement \| undefined \| Ref<HTMLElement \| undefined>` |

## Returns
`Ref<number>` (in px)


pleek91 (approved):
Can you update the root README? Just add this to the list. Looks great!

PR #58, opened 2022-05-25
Add useRouteQueryParam composition

pleek91:
# useRouteQueryParam
The `useRouteQueryParam` gives you a reactive value for a query param. So you can bind directly to `foo` in the route `example.com?foo=bar`

## Example
```typescript
import { useRouteQueryParam } from '@prefecthq/vue-compositions'

const foo = useRouteQueryParam('foo')
```

## Arguments
| Name         | Type                 | Default |
|--------------|----------------------|---------|
| param        | `string`             | None    |
| defaultValue | `string \| string[]` | `''`    |

## Returns
`string | string[]`


src/useRouteQueryParam/README.md
```diff
@@ -0,0 +1,18 @@
+# useRouteQueryParam
+The `useRouteQueryParam` gives you a reactive value for a query param. So you can bind directly to `foo` in the route `example.com?foo=bar`
+
+## Example
+```typescript
+import { useRouteQueryParam } from '@prefecthq/vue-compositions'
+
+const foo = useRouteQueryParam('foo')
```
stackoverfloweth, line 8: might be worth showing that you can also write to `foo`, which will push to router

stackoverfloweth (approved):
looks super useful!

src/useRouteQueryParam/README.md
pleek91: updated

PR #63, opened 2022-06-09
BugFix: useSubscription argument watcher incorrectly setting response

pleek91:
# Description
Previously the argument watcher would set the response of a new subscription that had a nullish response equal to the response of the initial subscription. The correct behavior is that it would set a nullish response equal to the previous response value.


src/useSubscription/useSubscription.ts
```diff
@@ -10,42 +10,38 @@ const defaultManager = new Manager()
...
+      }
 
-        response.unsubscribe()
+      subscriptionResponse.unsubscribe()
 
-        const newSubscription = manager.subscribe(action, argsWithDefault, options)
+      const newSubscription = manager.subscribe(action, argsWithDefault, options)
 
-        newSubscription.response.value ??= subscription.response.value
+      newSubscription.response.value ??= subscriptionResponse.response
```
pleek91, line 30: Everything in this file is renaming other than this line. This changed from `originalSubscription.response.value` to `subscriptionResponse.response`.

stackoverfloweth (approved):

PR #65, opened 2022-06-24
BugFix: fixed types to allow for single argument; added useRouteParam to vue-compositions

stackoverfloweth:
fixes this error
image

### DIFF
--- a/src/useRouteParam/README.md
+++ b/src/useRouteParam/README.md
@@ -1,0 +1,20 @@
+# useRouteParam
+The `useRouteParam` gives you a readonly reactive value for a query param. If you need the param to be writable, consider [useRouteQueryParam](https://github.com/PrefectHQ/vue-compositions/tree/main/src/useRouteQueryParam).
+
+## Example
+```typescript
+import { useRouteParam } from '@prefecthq/vue-compositions'
+
+const foo = useRouteParam('foo') // type is `string | string[]`
+const foo = useRouteParam('foo', 'default-string') // type is `string`
+const foo = useRouteParam('foo', ['default-array']) // type is `string[]`
+```
+
+## Arguments
+| Name         | Type                 | Default |
+|--------------|----------------------|---------|
+| param        | `string`             | None    |
+| defaultValue | `string \| string[]` | `undefined`    |
+
+## Returns
+`string | string[]`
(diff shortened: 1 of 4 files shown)


pleek91 (approved):

src/useRouteParam/README.md
```diff
@@ -0,0 +1,20 @@
+# useRouteParam
+The `useRouteParam` gives you a readonly reactive value for a query param. If you need the param to be writable, consider [useRouteQueryParam](https://github.com/PrefectHQ/vue-compositions/tree/main/src/useRouteQueryParam).
```
pleek91: It doesn't give you access to query params. It gives you the value of a route param. They're different. 

route.params[param]

Vs

route.query[param]

pleek91 (changes requested):

src/useRouteParam/README.md
stackoverfloweth: already fixed

src/useRouteParam/useRouteParam.ts
```diff
@@ -0,0 +1,28 @@
+/* eslint-disable no-redeclare */
+import { computed, Ref } from 'vue'
+import { useRoute } from 'vue-router'
+
+export function useRouteParam(param: string): Ref<string | string[]>
+export function useRouteParam(param: string, defaultValue: string): Ref<string>
+export function useRouteParam(param: string, defaultValue: string[]): Ref<string[]>
+export function useRouteParam(param: string, defaultValue?: string | string[]): Ref<string | string[]> {
```
pleek91, line 8: This should not support default values. They either exist in the url or they don't.

pleek91 (changes requested):

src/useRouteParam/useRouteParam.ts
pleek91: I wonder about the return type as well. Technically route params can be `string` or `string[]`. So technically this is correct. But I believe we've been using this assuming it's `string` and we'd have to do an additional type check every time we use this. 

I'd like to see a method for getting a param that returns `string` so we don't have to do additional type guarding. (maybe an additional method, or a `first` optional Boolean argument?)

pleek91 (approved):

PR #93, opened 2022-09-07
Feature: Use element rect

stackoverfloweth:
same as `useElementWidth` but includes `height`, `x`, and `y`

### DIFF
--- a/src/useElementRect/README.md
+++ b/src/useElementRect/README.md
@@ -1,0 +1,18 @@
+# useElementRect
+The `useElementRect` composition abstracts away the logic for tracking the client rect of an HTMLElement, even when the client rect changes on resize thanks to [useResizeObserver](https://github.com/PrefectHQ/vue-compositions/tree/main/src/useResizeObserver).
+
+## Example
+```typescript
+import { useElementRect } from '@prefecthq/vue-compositions'
+
+const templateRef = ref<HTMLElement | undefined>()
+const { height, width, x, y } = useElementRect(templateRef)
+```
+
+## Arguments
+| Name     | Type                              |
+|----------|-----------------------------------|
+| element | `HTMLElement \| undefined \| Ref<HTMLElement \| undefined>` |
+
+## Returns
+`Ref<number>` (in px)
--- a/src/useElementRect/README.md
+++ b/src/useElementRect/README.md
@@ -1,5 +1,5 @@
 # useElementRect
-The `useElementRect` composition abstracts away the logic for tracking the client rect of an HTMLElement, even when the client rect changes on resize thanks to [useResizeObserver](https://github.com/PrefectHQ/vue-compositions/tree/main/src/useResizeObserver).
+The `useElementRect` composition abstracts away the logic for tracking the client DOMRect of an HTMLElement, even when the client DOMRect changes on resize thanks to [useResizeObserver](https://github.com/PrefectHQ/vue-compositions/tree/main/src/useResizeObserver).
 
 ## Example
 ```typescript
@@ -15,4 +15,4 @@ const { height, width, x, y } = useElementRect(templateRef)
 | element | `HTMLElement \| undefined \| Ref<HTMLElement \| undefined>` |
 
 ## Returns
-`Ref<number>` (in px)
+`Ref<DOMRect>`
--- a/src/useElementRect/README.md
+++ b/src/useElementRect/README.md
@@ -6,7 +6,7 @@ The `useElementRect` composition abstracts away the logic for tracking the clien
 import { useElementRect } from '@prefecthq/vue-compositions'
 
 const templateRef = ref<HTMLElement | undefined>()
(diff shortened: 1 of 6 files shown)


pleek91:
@stackoverfloweth love this. Thoughts on just using the whole DOMRect from `getBoundingClientRect`? Rather than limiting it to width, height, x, and y

stackoverfloweth:
> @stackoverfloweth love this. Thoughts on just using the whole DOMRect from `getBoundingClientRect`? Rather than limiting it to width, height, x, and y

@pleek91 love that idea

src/useElementRect/README.md
```diff
@@ -0,0 +1,18 @@
+# useElementRect
+The `useElementRect` composition abstracts away the logic for tracking the client DOMRect of an HTMLElement, even when the client DOMRect changes on resize thanks to [useResizeObserver](https://github.com/PrefectHQ/vue-compositions/tree/main/src/useResizeObserver).
+
+## Example
+```typescript
+import { useElementRect } from '@prefecthq/vue-compositions'
+
+const templateRef = ref<HTMLElement | undefined>()
+const { height, width, x, y } = useElementRect(templateRef)
```
pleek91, line 9: Since this returns a `reactive` I don't think you can destructure this and retain reactivity

src/useElementRect/useElementRect.ts
```diff
@@ -0,0 +1,56 @@
...
+  width: Ref<number>,
+  x: Ref<number>,
+  y: Ref<number>,
+  left: Ref<number>,
+  top: Ref<number>,
+  right: Ref<number>,
+  bottom: Ref<number>,
+}
+
+export function useElementRect(element: HTMLElement | undefined | Ref<HTMLElement | undefined>): ElementRect {
```
pleek91, line 16: nit: I don't think there's a reason to accept undefined if its not a ref.

src/useElementRect/useElementRect.ts
```diff
@@ -0,0 +1,56 @@
...
+    clientRect.top.value = rect.top
+    clientRect.right.value = rect.right
+    clientRect.bottom.value = rect.bottom
+  }
+
+  const callback: UseResizeObserverCallback = function([entry]: ResizeObserverEntry[]) {
+    assignClientRect(entry.target.getBoundingClientRect())
+  }
+
+  const observer = useResizeObserver(callback)
```
pleek91, line 44: nit: I think this reads a little nicer and avoids having to define a const and import the callback type. 

```suggestion
  const observer = useResizeObserver(([entry]) => assignClientRect(entry.target.getBoundingClientRect()))
```

src/useElementRect/useElementRect.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  watchEffect(() => {
+    if (elementRef.value) {
+      assignClientRect(elementRef.value.getBoundingClientRect())
+
+      observer.disconnect()
+      observer.observe(elementRef)
+    }
+  })
```
pleek91, line 53: I think using `watchEffect` is a little risky here. `assignClientRect`, and `observer` all have reactive data that's being tracked when I think we only want this to run when `elementRef` changes.

src/useElementRect/useElementRect.ts
```diff
@@ -0,0 +1,56 @@
...
+  function assignClientRect(rect: DOMRect): void {
+    clientRect.height.value = rect.height
+    clientRect.width.value = rect.width
+    clientRect.x.value = rect.x
+    clientRect.y.value = rect.y
+    clientRect.left.value = rect.left
+    clientRect.top.value = rect.top
+    clientRect.right.value = rect.right
+    clientRect.bottom.value = rect.bottom
+  }
```
pleek91, line 38: nit: Accepting an element rather than a DOMRect means you don't have to call `getBoundingClientRect()` in the two places you call this method. 

```suggestion
  function assignClientRect(element: HTMLElement): void {
    const rect = element.getBoundingClientRect()
    
    clientRect.height.value = rect.height
    clientRect.width.value = rect.width
    clientRect.x.value = rect.x
    clientRect.y.value = rect.y
    clientRect.left.value = rect.left
    clientRect.top.value = rect.top
    clientRect.right.value = rect.right
    clientRect.bottom.value = rect.bottom
  }
```

src/useElementRect/useElementRect.ts
stackoverfloweth: I just copied useElementWidth, should I change it there as well?

pleek91: yeah I think so? Unless you can think of a reason to accept undefined. We only accept Ref<undefined> as a convenience since template refs are undefined during setup

src/useElementRect/README.md
pleek91: out dated

pleek91 (approved):

PR #96, opened 2022-09-11
Feature: added useChildrenAreWrapped composition

stackoverfloweth:
here is the problem I set out to solve
https://user-images.githubusercontent.com/6098901/189529322-63014562-1d90-4a7d-87d0-7e4b664de774.mov

the same behavior with this composition
https://user-images.githubusercontent.com/6098901/189529337-8874455b-b14e-419b-b0fc-ad5f5cc5ec64.mov

# useChildrenAreWrapped
The `useChildrenAreWrapped` composition returns a computed boolean that can be used in a component to react to the fact that children all fit within a single row, or they do not. This component uses [useElementRect](https://github.com/PrefectHQ/vue-compositions/tree/main/src/useElementRect) for sizing both, so keep in mind that if the container has padding, it will think there is more room for children than there actually is. 

## Example
```typescript
import { useChildrenAreWrapped } from '@prefecthq/vue-compositions'

const wrapped = useCongruentHeight(cogs, machine)
const classes = computed(() => ({
  container: {
    'p-container--wrapped': wrapped.value,
  },
}))
```

```css
.p-container {
  display: flex;
  flex-direction: row;
  justify-content: center;
}

.p-container--wrapped {
  flex-direction: column;
  justify-content: start;
}
```

## Arguments
| children | `HTMLElement[] \| Ref<HTMLElement[]` |
| container | `HTMLElement \| Ref<HTMLElement | undefined` |

## Returns
`ComputedRef<boolean>`


src/utilities/getComputedStyleValue.ts
```diff
@@ -0,0 +1,22 @@
+// https://stackoverflow.com/a/62557418/3511012
+function globalExists(varName: string): boolean {
+  const globalEval = eval
+  try {
+    globalEval(varName)
+    return true
+  } catch {
+    return false
+  }
+}
```
stackoverfloweth, line 10: not a fan of this pattern, but this is carried over from our work in saving `crypto` from blowing up when running in node (like in tests). Note this is only used to ensure `window` actually exists and is not exported

src/utilities/getComputedStyleValue.ts
```diff
@@ -0,0 +1,22 @@
...
+  if (globalExists('window')) {
+    try {
+      return window
+        .getComputedStyle(element, null)
+        .getPropertyValue(property)
+    } catch {
+    // do nothing
+    }
+  }
+}
```
stackoverfloweth, line 22: I considered using this to get actual container size minus padding, but it ended up pretty complex. In my use case I just make sure the parent of the children didn't have any padding.

stackoverfloweth:
here is the problem I set out to solve
https://user-images.githubusercontent.com/6098901/189529322-63014562-1d90-4a7d-87d0-7e4b664de774.mov

the same behavior with this composition
https://user-images.githubusercontent.com/6098901/189529337-8874455b-b14e-419b-b0fc-ad5f5cc5ec64.mov

src/utilities/getComputedStyleValue.ts
```diff
@@ -0,0 +1,22 @@
...
+  const globalEval = eval
+  try {
+    globalEval(varName)
+    return true
+  } catch {
+    return false
+  }
+}
+
+export function getComputedStyleValue(element: Element, property: string): string | undefined {
```
pleek91, line 12: Can we make this another composition? `useElementComputedStyle` That could be used like this could be really useful. 

```typescript
const { paddingLeft, paddingRight } = useElementComputedStyle(elementRef)
```

src/useChildrenAreWrapped/useChildrenAreWrapped.ts
```diff
@@ -0,0 +1,30 @@
+import { computed, ComputedRef, ref, Ref } from 'vue'
+import { useElementRect } from '@/useElementRect/useElementRect'
+import { getComputedStyleValue } from '@/utilities/getComputedStyleValue'
+import { getNumberFromPx } from '@/utilities/getNumberFromPx'
+
+export function useChildrenAreWrapped(children: HTMLElement[] | Ref<HTMLElement[]>, container: HTMLElement | Ref<HTMLElement | undefined>): ComputedRef<boolean> {
+  const childrenRef = ref(children)
+  const containerRef = ref(container)
+
+  const { width } = useElementRect(containerRef)
```
pleek91, line 10: I think you can include padding here without to much effort no?

src/useChildrenAreWrapped/useChildrenAreWrapped.ts
```diff
@@ -0,0 +1,30 @@
...
+
+    return getNumberFromPx(gapInPx)
+  }
+
+  const childrenWidthSum = computed(() => {
+    const rects = childrenRef.value.map(useElementRect)
+    const gap = getContainerGap()
+
+    return rects.reduce((sum, { width }) => sum += width.value + gap, 0)
+  })
```
pleek91, line 27: I'm not sure this technique is solid. Elements can have margins that would cause them to wrap sooner than this calculation would determine. Also depending on the box model borders could also effect when the elements would wrap.

src/utilities/getComputedStyleValue.ts
```diff
@@ -0,0 +1,22 @@
+// https://stackoverflow.com/a/62557418/3511012
```
pleek91, line 1: If we're introducing generic utilities in this repository can we keep the naming structure we've used in our other projects? `utilities/styles.ts` for example.

src/utilities/getComputedStyleValue.ts
stackoverfloweth: I could. I put it here to make it even harder to accidentally get exported and used somewhere that a user might provide the eval value.

src/utilities/getNumberFromPx.ts
```diff
@@ -0,0 +1,9 @@
+export function getNumberFromPx(sizeWithPx: string | undefined): number {
```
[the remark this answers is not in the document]
stackoverfloweth: good call, just overcomplicated it for no reason

src/utilities/getComputedStyleValue.ts
stackoverfloweth: love this

src/useChildrenAreWrapped/useChildrenAreWrapped.ts
stackoverfloweth: good call again

src/useChildrenAreWrapped/useChildrenAreWrapped.ts
stackoverfloweth: might not work in 100% of circumstances but it definitely solved my problem 🤷

src/useComputedStyle/useComputedStyle.ts
```diff
@@ -0,0 +1,16 @@
+/* eslint-disable no-redeclare */
+import { reactive, ref, Ref, ToRefs, toRefs, watchEffect } from 'vue'
+import { globalExists } from '@/utilities/global'
+
+export function useComputedStyle(element: Element | Ref<Element | undefined>): ToRefs<Partial<CSSStyleDeclaration>> {
```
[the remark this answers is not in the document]
pleek91: More thoughts. A ResizeObserver and a MutationObserver would catch MOST css changes. However its still possible for an elements computedStyle to be changed without triggering either observer. 

For instance a style is inherited or changed without changing the elements classes. If that change doesn't change the size of the element. Neither observer will be triggered.

src/useComputedStyle/useComputedStyle.ts
```diff
@@ -0,0 +1,16 @@
...
+  const elementRef = ref(element)
+  const style = reactive({})
+
+  watchEffect(() => {
+    if (elementRef.value && globalExists('window')) {
+      Object.assign(style, window.getComputedStyle(elementRef.value, null))
+    }
+  })
+
+  return toRefs(style)
```
pleek91, line 15: Nice use of `toRefs`

src/useChildrenAreWrapped/useChildrenAreWrapped.ts
```diff
@@ -0,0 +1,27 @@
+import { computed, ComputedRef, ref, Ref } from 'vue'
+import { useComputedStyle } from '@/useComputedStyle/useComputedStyle'
+import { useElementRect } from '@/useElementRect/useElementRect'
+
+export function useChildrenAreWrapped(children: HTMLElement[] | Ref<HTMLElement[]>, container: HTMLElement | Ref<HTMLElement | undefined>): ComputedRef<boolean> {
+  const childrenRef = ref(children)
+  const containerRef = ref(container)
+
+  const { width } = useElementRect(containerRef)
+  const { gap, paddingLeft, paddingRight } = useComputedStyle(container)
```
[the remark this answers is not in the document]
stackoverfloweth: twas a great suggestion

src/useComputedStyle/useComputedStyle.ts
```diff
@@ -0,0 +1,41 @@
...
+  watchEffect(() => {
+    if (elementRef.value && globalExists('window')) {
+      Object.assign(style, window.getComputedStyle(elementRef.value, null))
+
+      mutationObserver.disconnect()
+      mutationObserver.observe(elementRef)
+      resizeObserver.disconnect()
+      resizeObserver.observe(elementRef)
+    }
+  })
```
pleek91, line 20: I think all of this we only want to happen when `elementRef` changes. I'd convert this to a standard watcher to avoid unnecessary/unintended execution of this callback to do reactivity changes.

src/useMutationObserver/README.md
```diff
@@ -0,0 +1,22 @@
+# useMutationObserver
```
pleek91, line 1: ❤️

src/useResizeObserver/useResizeObserver.ts
```diff
@@ -1,17 +1,17 @@
...
-  check: (element: HTMLElement | Ref<HTMLElement | undefined>) => void,
+  check: (element: Element | Ref<Element | undefined>) => void,
 }
 
 export type UseResizeObserverCallback = (entries: ResizeObserverEntry[]) => void
 
 export function useResizeObserver(callback: UseResizeObserverCallback): UseResizeObserverResponse {
 
-  let intersectionObserver: ResizeObserver | null = null
+  let resizeObserver: ResizeObserver | null = null
```
pleek91, line 14: 😅 thanks 😂

stackoverfloweth: I only caught it because I did the same thing 😅

src/useComputedStyle/useComputedStyle.ts
stackoverfloweth: I think is pretty safe? It only executes if `elementRef.value` is truthy, so first time through when it's probably `undefined`, nothing will happen

pleek91: Right but once's it's truthy it'll run again and then I think it'll start tracking anything reactive inside that if? Could be wrong

src/useChildrenAreWrapped/useChildrenAreWrapped.ts
pleek91: Might be worth noting its limitations then if we know there are limitations

stackoverfloweth: I think I've accounted for such limitations, and updated readme. Take a look at let me know what you think

pleek91 (approved):

PR #103, opened 2022-09-13
Chore: updated useChildrenAreWrapped to not call compositions in callbacks

stackoverfloweth:
previously, `useChildrenAreWrapped` called other compositions from a computed callback. It is not permitted in vue to call compositions from outside of the setup context.

this PR introduces a simplification of that logic to compute total children width only when the container changes, rather than also watching for changes within any individual child.


src/useChildrenAreWrapped/useChildrenAreWrapped.ts
```diff
@@ -1,36 +1,33 @@
...
 
-    return childrenSize.reduce((sum, { rect, style }) => {
-      const margin = getPxValue(style.marginLeft) + getPxValue(style.marginRight)
-      const border = getPxValue(style.borderTopWidth) + getPxValue(style.borderBottomWidth)
-      const containerGap = getPxValue(gap)
+    return styles.reduce((sum, style) => {
+      if (style) {
+        sum += style.width
+        sum += style.marginLeft + style.marginRight
+        sum += style.borderTopWidth + style.borderBottomWidth
```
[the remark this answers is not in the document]
stackoverfloweth: nice, good call

src/useComputedStyle/useComputedStyle.ts
```diff
@@ -2,29 +2,33 @@
...
-import { globalExists } from '../utilities/global'
+import { DocumentDoesNotExist } from './documentDoesNotExist'
+import { globalExists } from '@/utilities/global'
+import { ComputedStyleRecord, getComputedStyleRecord } from '@/utilities/window'
 
-export function useComputedStyle(element: Element | Ref<Element | undefined>): ToRefs<Partial<CSSStyleDeclaration>> {
+export function useComputedStyle(element: Element | Ref<Element | undefined>): ToRefs<ComputedStyleRecord> {
   const elementRef = ref(element)
-  const style = reactive({})
+  const style = reactive(getComputedStyleRecord(elementRef.value ?? getDefaultElement())!)
```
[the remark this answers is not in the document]
stackoverfloweth: I did this so that we can basically new up a `CSSStyleDeclaration` so our return type is consistent. Otherwise I'd have to return a partial or undefined

src/useChildrenAreWrapped/useChildrenAreWrapped.ts
```diff
@@ -1,36 +1,33 @@
...
   const { width } = useElementRect(containerRef)
-  const { gap, paddingLeft, paddingRight } = useComputedStyle(container)
+  const { columnGap, paddingLeft, paddingRight } = useComputedStyle(container)
 
-  const childrenWidthSum = computed(() => {
-    const childrenSize = childrenRef.value.map(element => ({
-      rect: useElementRect(element),
-      style: useComputedStyle(element),
-    }))
+  function getChildrenWidth(): number {
```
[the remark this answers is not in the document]
stackoverfloweth: agreed

src/useComputedStyle/useComputedStyle.ts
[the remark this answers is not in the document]
pleek91: We haven't done this before. But what if we overloaded the composition so if you give us a ref we'll return the declaration or null?

src/utilities/arrays.ts
```diff
@@ -0,0 +1,28 @@
...
+  return array.every(value => typeof value === 'string')
+}
+
+export function reduceToRecord<S, K extends string | number | symbol, V>(array: S[], key: (item: S) => K, value: (item: S) => V): Record<K, V> {
+  return array.reduce((reduced, item) => {
+    reduced[key(item)] = value(item)
+
+    return reduced
+  // eslint-disable-next-line @typescript-eslint/prefer-reduce-type-parameter
+  }, {} as Record<K, V>)
```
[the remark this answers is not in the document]
pleek91: huh, swear I did that yesterday to fix that. Idk, ignore.

src/useComputedStyle/useComputedStyle.ts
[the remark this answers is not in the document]
stackoverfloweth: I agree that this is a bit unique. Unfortunately I think this is an unintended consequence of the decision to return `toRefs` of a `reactive`. If my reactive starts as `{}`, and thus the return value is 

```ts
const { columnGap } = useComputedStyle(element)

// typeof columnGap === undefined <- not good
// different from typeof columnGap.value === undefined <- would probably work fine
```

then, even after columnGap theoretically gets populated, it's not reactive like it would be if only `.value` was undefined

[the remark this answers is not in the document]
pleek91: Ah right. Well I'm not a fan of the span thing but I also don't have any better ideas.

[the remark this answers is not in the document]
stackoverfloweth: I could move away from `toRefs` and just return `Ref<CSSStyleDeclaration | undefined>`? That would mean we can't do 

```ts
const { columnGap, paddingRight } = useComputedStyle(element)
```

instead we would have to do 

```ts
const styles = useComputedStyle(element)
// styles.value?.columnGap
```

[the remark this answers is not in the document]
stackoverfloweth: 3rd option, define my own class that looks like `CSSStyleDeclaration` with properties like `width`, `paddingRight`, etc, but can be newed up without being given an element. One positive outcome of that approach is we could type the things that should be numbers as number instead of all being strings.

what's your favorite?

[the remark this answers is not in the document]
stackoverfloweth: here's what that class looks like with only the properties I need. There are approx 600 total properties on `CSSStyleDeclaration` 😬 

```ts
export class ComputedStyleRecord {
  public width: number = 0
  public paddingLeft: number = 0
  public paddingRight: number = 0
  public marginLeft: number = 0
  public marginRight: number = 0
  public columnGap: number = 0
  public boxSizing: string = ''

  public constructor(cssStyleDeclaration?: CSSStyleDeclaration) {
    if (cssStyleDeclaration) {
      this.width = parseInt(cssStyleDeclaration.getPropertyValue('padding-left'))
      this.paddingLeft = parseInt(cssStyleDeclaration.getPropertyValue('padding-left'))
      this.paddingRight = parseInt(cssStyleDeclaration.getPropertyValue('padding-right'))
      this.marginLeft = parseInt(cssStyleDeclaration.getPropertyValue('margin-left'))
      this.marginRight = parseInt(cssStyleDeclaration.getPropertyValue('margin-right'))
      this.columnGap = parseInt(cssStyleDeclaration.getPropertyValue('column-gap'))
      this.boxSizing = cssStyleDeclaration.getPropertyValue('box-sizing')
    }
  }
}
```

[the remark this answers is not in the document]
pleek91: Interesting and useful to be able to do some type casting of numbers vs strings. Since most of the time we want the number value not the px value. 

That's, a, lot, of properties though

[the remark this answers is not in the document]
stackoverfloweth: my vote is #2, just return `Ref<CSSStyleDeclaration | undefined>`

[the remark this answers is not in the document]
pleek91: Seems the simplest. 👍

src/utilities/window.ts
```diff
@@ -0,0 +1,22 @@
...
+
+  const computedStyle = window.getComputedStyle(element)
+
+  return reduceToRecord(Object.keys(computedStyle),
+    key => key as keyof CSSStyleDeclaration,
+    key => {
+      const kebab = kebabCase(key)
+
+      return computedStyle.getPropertyValue(kebab)
+    },
```
pleek91, line 20: What is this doing? Doesn't getComputedStyle already return an object with all the property styles?

stackoverfloweth: I don't full remember. I know CSSStyleDeclaration is a class and I was seeing weird things like keys both as kebab and as pascal. Either way, the ref change seems to have removed the need for it. Good catch again

src/utilities/window.ts
```diff
@@ -0,0 +1,9 @@
+import { globalExists } from './global'
+
+export function getComputedStyleRecord(element: Element | undefined): CSSStyleDeclaration | undefined {
```
pleek91, line 3: Is there a particular reason for the Record suffix? We don't generally add the return type to our methods.

stackoverfloweth: Just left over from when i was changing the value into a record. I'll fix

stackoverfloweth, line 3:
```suggestion
export function getWindowComputedStyle(element: Element | undefined): CSSStyleDeclaration | undefined {
```

stackoverfloweth: Although it should be different. Otherwise vscode will assume we want the method from `window`

src/useComputedStyle/useComputedStyle.ts
```diff
@@ -1,30 +1,33 @@
-/* eslint-disable no-redeclare */
-import { reactive, ref, Ref, ToRefs, toRefs, watch } from 'vue'
+import { ref, Ref, watch } from 'vue'
 import { useMutationObserver } from '../useMutationObserver/useMutationObserver'
 import { useResizeObserver } from '../useResizeObserver/useResizeObserver'
-import { globalExists } from '../utilities/global'
+import { getComputedStyleRecord } from '@/utilities/window'
```
stackoverfloweth, line 4:
```suggestion
import { getWindowComputedStyle } from '@/utilities/window'
```

src/useComputedStyle/useComputedStyle.ts
```diff
@@ -1,30 +1,33 @@
...
 
-export function useComputedStyle(element: Element | Ref<Element | undefined>): ToRefs<Partial<CSSStyleDeclaration>> {
+export function useComputedStyle(element: Element | Ref<Element | undefined>): Ref<CSSStyleDeclaration | undefined> {
   const elementRef = ref(element)
-  const style = reactive({})
+  const initialStyle = getComputedStyleRecord(elementRef.value)
```
stackoverfloweth, line 8:
```suggestion
  const initialStyle = getWindowComputedStyle(elementRef.value)
```

src/useComputedStyle/useComputedStyle.ts
```diff
@@ -1,30 +1,33 @@
...
+      updateStyleRef(entry.target)
     }
   }
-  const mutationObserver = useMutationObserver(handleMutation)
 
-  function handleResize([entry]: ResizeObserverEntry[]): void {
-    if (globalExists('window')) {
-      Object.assign(style, window.getComputedStyle(entry.target, null))
+  function updateStyleRef(element: Element): void {
+    const computedStyleRecord = getComputedStyleRecord(element)
```
stackoverfloweth, line 18:
```suggestion
    const computedStyleRecord = getWindowComputedStyle(element)
```

src/useComputedStyle/useComputedStyle.ts
```diff
@@ -1,30 +1,33 @@
...
+
+    if (computedStyleRecord) {
+      style.value = computedStyleRecord
```
stackoverfloweth, line 21:
```suggestion
    if (computedStyle) {
      style.value = computedStyle
```

src/useChildrenAreWrapped/useChildrenAreWrapped.ts
```diff
@@ -1,36 +1,48 @@
 import { computed, ComputedRef, ref, Ref } from 'vue'
 import { useComputedStyle } from '../useComputedStyle/useComputedStyle'
 import { useElementRect } from '../useElementRect/useElementRect'
+import { getComputedStyleRecord } from '@/utilities/window'
```
stackoverfloweth, line 4:
```suggestion
import { getWindowComputedStyle } from '@/utilities/window'
```

src/useChildrenAreWrapped/useChildrenAreWrapped.ts
```diff
@@ -1,36 +1,48 @@
...
 
-  return computed(() => childrenWidthSum.value > containerWidthWithoutPadding.value)
+  return parseInt(style)
 }
 
-function getPxValue(property: Ref<string | undefined> | undefined): number {
-  return parseInt(property?.value ?? '0')
+function getChildrenWidth(elements: Element[], gap: number): number {
+  return elements
+    .map(getComputedStyleRecord)
```
stackoverfloweth, line 35:
```suggestion
    .map(getWindowComputedStyle)
```

src/useComputedStyle/useComputedStyle.ts
```diff
@@ -1,30 +1,33 @@
...
+      updateStyleRef(entry.target)
     }
   }
-  const mutationObserver = useMutationObserver(handleMutation)
 
-  function handleResize([entry]: ResizeObserverEntry[]): void {
-    if (globalExists('window')) {
-      Object.assign(style, window.getComputedStyle(entry.target, null))
+  function updateStyleRef(element: Element): void {
+    const computedStyleRecord = getWindowComputedStyle(element)
```
stackoverfloweth, line 18:
```suggestion
    const computedStyle = getWindowComputedStyle(element)
```

src/utilities/window.ts
pleek91: Does it really? That's dumb.

stackoverfloweth: it looks like when I import it, it does override the window. But the fact that it can appear to work when not imported because it's actually using the version from `window` makes me feel like a slightly different name is a better route

https://user-images.githubusercontent.com/6098901/190163096-b3f6a3b8-9906-41d3-bfc5-d19871ca6218.mov

pleek91 (approved):

PR #104, opened 2022-09-14
BugFix: removed alias imports (@), removed option from tsconfig

stackoverfloweth:
Builds fail when we use imports like `import { thing } from '@/utilities/thing`. 
My guess is that it's because vue-compositions isn't transpiled? Until it is, I think we should remove the option from tsconfig

image

image


pleek91 (approved):
We really need to publish this with a build. Maybe I can work on that this week

PR #108, opened 2022-10-06
Feature: useStorage

pleek91:
# useStorage
The `useStorage` composition is a reactive wrapper around localStorage and sessionStorage. 

## Example
```typescript
import { useStorage } from '@prefecthq/vue-compositions'

const { values: foo } = useStorage('local', 'foo') // type is `unknown`
const { values: bar } = useStorage<boolean>('local', 'bar') // type is `boolean`
const { values: bas } = useStorage('local', 'bas', false) // type 'boolean'
```
Two additional compositions are exported as a convenience for each type of storage. `useLocalStorage` and `useSessionStorage`. 
```typescript
import { useLocalStorage, useSessionStorage } from '@prefect/vue-compositions'

// this
const { values: foo } = useLocalStorage('foo')

// is the same as this
const { values: foo } = useStorage('local', 'foo')

// and this
const { values: foo } = useSessionStorage('foo')

// is the same as this
const { values: foo } = useStorage('session', 'foo')

```

## Arguments
| Name         | Type               | Default | Required
|--------------|--------------------|---------|-----------|
| type         | `local \| session` | None    | true
| key          | `string`           | None    | true
| defaultValue | `any`              | None    | false

## Returns
```typescript
type UseStorage<T> = {
  value: Ref<T>,
  initialValue: T,
  remove: () => void,
  set: (value: NonNullable<T>) => void,
}
```


src/useStorage/storage.ts
```diff
@@ -0,0 +1,92 @@
...
+      return localStorage
+    }
+
+    if (this.type === 'session' && globalExists('sessionStorage')) {
+      return sessionStorage
+    }
+
+    return null
+  }
+}
```
stackoverfloweth, line 92: I'm surprised to see this is not a computed getter?

src/useStorage/storage.ts
```diff
@@ -0,0 +1,92 @@
+import { globalExists } from '../utilities/global'
+
+export type StorageType = 'session' | 'local'
+
+export class StorageManager {
```
stackoverfloweth, line 5: this is 100% cool as is, just sharing what my approach would have been in case you think it's groovy. 

what if instead of defining a single class that changes functionality of each method based on a single property `type`, you created a separate class for localStorage and sessionStorage that both implemented `StorageManager` or something like that. StorageManager could have protected methods to abstract away the JSON work so each implementation doesn't have to duplicate that logic.

src/useStorage/useStorage.ts
```diff
@@ -0,0 +1,52 @@
+/* eslint-disable no-redeclare */
```
stackoverfloweth, line 1: we should really kill that eslint rule

src/useStorage/useStorage.ts
```diff
@@ -0,0 +1,52 @@
...
+  remove: () => void,
+  set: (value: NonNullable<T>) => void,
+}
+
+export function useStorage<T>(type: StorageType, key: string): UseStorage<T | null>
+export function useStorage<T>(type: StorageType, key: string, defaultValue: T): UseStorage<T>
+export function useStorage<T>(type: StorageType, key: string, defaultValue: T | null = null): UseStorage<T | null> {
+  const storage = new StorageManager(type)
+  const initialValue = storage.get(key, defaultValue)
+  const data = ref(initialValue) as Ref<T | null>
```
stackoverfloweth, line 17: I think the type that vue is suggesting here is actually more accurate. If I use `useStorage` and my `T` is `Ref<string>`, when you set `data = ref<T>` it will unref `T` to `string`. 

That means your `UseStorage` type should be
```ts
type UseStorage<T> = {
  value: Ref<UnwrapRef<T>>,
  initialValue: T,
  remove: () => void,
  set: (value: NonNullable<UnwrapRef<T>>) => void,
}
```

without this change, if `T` is a ref, ts will mistakingly expect you to be nesting your refs
```ts
const initVal = ref('poop')
const test = useSessionStorage('typescript', initVal)

test.set('sweet') // Argument of type 'string' is not assignable to parameter of type 'Ref<string>'.ts(2345)
```

with the Unref<T> change
```ts
const initVal = ref('poop')
const test = useSessionStorage('typescript', initVal)

test.set('sweet') // works
```

src/useStorage/README.md
```diff
@@ -0,0 +1,45 @@
+# useStorage
```
stackoverfloweth, line 1: I just want to verify, if someone were to manually change a value in storage with `localStorage.set()`, this composition's value ref would NOT know about it, correct? 

Perhaps we should make sure that's noted in the readme?

stackoverfloweth (commented):
Thanks @pleek91, another great composition for the library!

src/useStorage/storage.ts
pleek91: Both localStorage and sessionStorage have the exact same api. So the two classes would be identical except for calling `localStorage` or `sessionStorage`. Considering that's the only difference between the two storage types. Once class is favorable over three IMO.

src/useStorage/storage.ts
pleek91: I suppose it probably could be. But my intent was to not compute this until it's needed.

src/useStorage/useStorage.ts
pleek91: I think its a conflicting js vs ts rule. We really need to rework that config...

src/useStorage/README.md
pleek91: Correct, but that is also expected. There is no browser api for watching/listening to changes to storage. If you don't go through the composition there shouldn't be any expectation that your update will be reactive.

src/useStorage/useStorage.ts
pleek91: Hmm I see the issue, but changing that type I think will cause other issues? Let's look at this sync

pleek91: Update 👍

stackoverfloweth (approved):

PR #116, opened 2022-10-25
BugFix: useStorage better support transferring old storage values

stackoverfloweth:
if useStorage is used on previously saved value, this will prevent console error and allow safe transition

for example, color mode was previously saved to localStorage with simple

```ts
localStorage.setItem(colorModeLocalStorageKey, mode)
// mode is string, like 'achromatopsia'
```

if your user had 'achromatopsia' in localStorage, and you try updating to useStorage

```ts
const { value } = useStorage(colorModeLocalStorageKey)
```

you'll get the error

> Uncaught SyntaxError: Unexpected token 'a', "achromatopsia" is not valid JSON

with this PR, you'll be able to write 

```ts
const nonJsonVersion = localStorage.getItem(colorModeLocalStorageKey)
const defaultValue = isColorMode(nonJsonVersion) ? nonJsonVersion : null
```

which safely transfers the `nonJsonVersion` in

### DIFF
--- a/src/useStorage/storage.ts
+++ b/src/useStorage/storage.ts
@@ -37,6 +37,7 @@ export class StorageManager {
     try {
       return JSON.parse(value) as T
     } catch {
+      console.error(`Unable to parse current value for key ${key}, returning default instead`)
       return defaultValue
     }
   }
--- a/src/useStorage/storage.ts
+++ b/src/useStorage/storage.ts
@@ -34,7 +34,11 @@ export class StorageManager {
       return defaultValue
     }
 
-    return JSON.parse(value) as T
+    try {
+      return JSON.parse(value) as T
+    } catch {
+      return defaultValue
+    }
   }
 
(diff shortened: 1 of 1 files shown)


src/useStorage/storage.ts
```diff
@@ -34,7 +34,11 @@ export class StorageManager {
       return defaultValue
     }
 
-    return JSON.parse(value) as T
+    try {
+      return JSON.parse(value) as T
+    } catch {
+      return defaultValue
+    }
```
pleek91, line 41: @stackoverfloweth swallowing this error seems like a bad idea. 

I'm also wondering if we shouldn't return `value ?? defaultValue`? That way we shouldn't need to do anything special to support migrating old storage values.

stackoverfloweth: @pleek91 I don't follow, can you show me an example of what you're thinking?

pleek91 (approved):

PR #122, opened 2022-11-09
Feature: Create useVisibilityObserver

marichka-offen:
This PR adds the `useVisibilityObserver` composition that is based on `useIntersectionObserver` composition to abstract away the logic for creating and connecting an IntersectionObserver and returns a boolean value if the target element is visible

### DIFF
--- a/src/useVisibilityObserver/useVisibilityObserver.ts
+++ b/src/useVisibilityObserver/useVisibilityObserver.ts
@@ -6,20 +6,24 @@ export type UseVisibilityObserverResponse = {
   disconnect: () => void,
 }
 
-export function useVisibilityObserver(element: Ref<HTMLElement | undefined>, disconnectWhenVisible: boolean = false, options: UseIntersectionObserverOptions = {}): UseVisibilityObserverResponse {
+export type UseVisibilityObserverOptions = UseIntersectionObserverOptions & { disconnectWhenVisible: boolean }
+
+export function useVisibilityObserver(element: Ref<HTMLElement | undefined>, options: UseVisibilityObserverOptions = { disconnectWhenVisible: false } ): UseVisibilityObserverResponse {
   const visible = ref(false)
 
   function intersect(entries: IntersectionObserverEntry[]): void {
     entries.forEach(entry => {
-        visible.value = entry.isIntersecting
+      visible.value = entry.isIntersecting
 
-        if (disconnectWhenVisible) {
-          disconnect()
-        }
+      if (options.disconnectWhenVisible && entry.isIntersecting) {
+        disconnect()
+      }
     })
   }
 
-  const { observe, disconnect } = useIntersectionObserver(intersect, options)
+  const { root, rootMargin, threshold } = options
+
+  const { observe, disconnect } = useIntersectionObserver(intersect, { root, rootMargin, threshold })
 
   onMounted(() => {
     observe(element)
--- a/src/useVisibilityObserver/useVisibilityObserver.ts
+++ b/src/useVisibilityObserver/useVisibilityObserver.ts
@@ -1,24 +1,29 @@
 import { onMounted, ref, Ref } from 'vue'
-import { useIntersectionObserver } from '@/useIntersectionObserver'
+import { useIntersectionObserver, UseIntersectionObserverOptions } from '@/useIntersectionObserver'
 
-export function useVisibilityObserver(element: Ref<HTMLElement | undefined>): Ref<boolean> {
+export type UseVisibilityObserverResponse = {
+  visible: Ref<boolean>,
+  disconnect: () => void,
+}
+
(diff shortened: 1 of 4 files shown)


src/useVisibilityObserver/useVisibilityObserver.ts
```diff
@@ -0,0 +1,24 @@
...
+    })
+  }
+
+  const { observe, disconnect } = useIntersectionObserver(intersect)
+
+  onMounted(() => {
+    observe(element)
+  })
+
+  return visible
```
pleek91, line 22: We should return an object of `{ visible, disconnect }` in case the dev wants to disconnect manually.

src/useVisibilityObserver/useVisibilityObserver.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export function useVisibilityObserver(element: Ref<HTMLElement | undefined>): Ref<boolean> {
+  const visible = ref(false)
+
+  function intersect(entries: IntersectionObserverEntry[]): void {
+    entries.forEach(entry => {
+      if (entry.isIntersecting) {
+        visible.value = true
+        disconnect()
+      }
```
pleek91, line 12: As a generic composition I think we shouldn't auto disconnect and we should track visibility. 

```suggestion
      visible.value = entry.isIntersecting
```

src/useVisibilityObserver/useVisibilityObserver.ts
```diff
@@ -0,0 +1,24 @@
+import { onMounted, ref, Ref } from 'vue'
+import { useIntersectionObserver } from '@/useIntersectionObserver'
+
+export function useVisibilityObserver(element: Ref<HTMLElement | undefined>): Ref<boolean> {
```
pleek91, line 4: We should expose the `options` from `useIntersectionObserver` here as an optional second argument. So a developer can customize the root, threshold, etc.

src/useVisibilityObserver/useVisibilityObserver.ts
pleek91: Since its a common use case though maybe we can add a `disconnectWhenVisible: boolean` option that does do the disconnect automatically.

pleek91 (commented):
Thank you! Couple thoughts about making this a little more generic but this will be so nice to use in our projects

src/useVisibilityObserver/useVisibilityObserver.ts
```diff
@@ -0,0 +1,29 @@
+import { onMounted, ref, Ref } from 'vue'
+import { useIntersectionObserver, UseIntersectionObserverOptions } from '@/useIntersectionObserver'
+
+export type UseVisibilityObserverResponse = {
+  visible: Ref<boolean>,
+  disconnect: () => void,
+}
+
+export function useVisibilityObserver(element: Ref<HTMLElement | undefined>, disconnectWhenVisible: boolean = false, options: UseIntersectionObserverOptions = {}): UseVisibilityObserverResponse {
```
pleek91, line 9: Can we combine `disconnectWhenVisible` with the other `options` so we accept only two arguments? `element` and `options`.

src/useVisibilityObserver/useVisibilityObserver.ts
```diff
@@ -0,0 +1,29 @@
...
+  const visible = ref(false)
+
+  function intersect(entries: IntersectionObserverEntry[]): void {
+    entries.forEach(entry => {
+        visible.value = entry.isIntersecting
+
+        if (disconnectWhenVisible) {
```
pleek91, line 16: We'll need to check to make sure `entry.isIntersecting` here as well I believe.

src/useVisibilityObserver/useVisibilityObserver.ts
```diff
@@ -0,0 +1,33 @@
...
+
+      if (options.disconnectWhenVisible && entry.isIntersecting) {
+        disconnect()
+      }
+    })
+  }
+
+  const { root, rootMargin, threshold } = options
+
+  const { observe, disconnect } = useIntersectionObserver(intersect, { root, rootMargin, threshold })
```
pleek91, line 26: Slight change to how options are used. This will prevent this composition from needing to be updated if the options for `useIntersectionObserver` change. 

```suggestion
  const visible = ref(false)
  const { disconnectWhenVisible, ...intersectionObserverOptions } = options

  function intersect(entries: IntersectionObserverEntry[]): void {
    entries.forEach(entry => {
      visible.value = entry.isIntersecting

      if (disconnectWhenVisible && entry.isIntersecting) {
        disconnect()
      }
    })
  }

  const { observe, disconnect } = useIntersectionObserver(intersect, intersectionObserverOptions)
```

pleek91:
@marichka-offen thank you so much! One very last thing (I promise). Can you add a link to this new composition in the rood README.md?

marichka-offen:
Ofc!

marichka-offen:
@pleek91 README is updated

pleek91 (approved):
THANK YOU!

PR #128, opened 2022-11-21
Chore: Add a build

### LINKED ISSUE #16: Create a build
Currently there isn't an actual build. The src is published as is and compositions can only be used in typescript projects. Need to create a dist package for native js

pleek91:
# Description
Closes #16

### DIFF
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -1,0 +1,29 @@
+import { resolve } from 'path'
+
+// https://vitejs.dev/config/
+export default {
+  resolve: {
+    alias: [
+      {
+        find: '@',
+        replacement: resolve(__dirname, './src'),
+      },
+    ],
+  },
+  build: {
+    emptyOutDir: false,
+    lib: {
+      entry: resolve(__dirname, 'src/index.ts'),
+      name: 'vue-compositions',
+    },
+    rollupOptions: {
+      external: ['vue', 'vue-router'],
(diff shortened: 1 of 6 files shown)


vite.config.ts
```diff
@@ -0,0 +1,29 @@
...
+  resolve: {
+    alias: [
+      {
+        find: '@',
+        replacement: resolve(__dirname, './src'),
+      },
+    ],
+  },
+  build: {
+    emptyOutDir: false,
```
znicholasbrown, line 14: I think you can omit this field (it defaults to true) unless there's a reason you wouldn't want to remove files in the `/dist` folder whenever you're making a build?

znicholasbrown: Ah just kidding, that's so your type files aren't overwritten

pleek91: That's what we're doing in the other projects. I remember we added it for some reason. But I don't remember why. Seems like we shouldn't need it though 🤔 I'll remove.

znicholasbrown: No we for sure need it since the `tsc` command you're chaining puts files in `/dist/types`

pleek91: jk its back! 😂

znicholasbrown (approved):

znicholasbrown:
Looks good to me @pleek91

PR #137, opened 2022-12-08
Feature: usePatchRef composition

pleek91:
Anyone have a better name for this? 

# usePatchRef
The `usePatchRef` abstracts away the get/set pattern for updating a computed object that has a setter. Normally you would need a get/set for each property you want to reactively set. Using the `usePatchRef` composition makes this simpler.

## Example
```typescript
import { usePatch } from '@prefecthq/vue-compositions'
import { computed } from 'vue'

type ValueType = Record<'myKey' | 'myOtherKey', string>

const props = defineProps<
  value: ValueType
>()

const emit = defineEmits<{
  (event: 'update:value', value: Record<'myKey' | 'myOtherKey', string>)
}>()

const internalValue = computed({
  get() {
    return props.value
  },
  set(value: Record<'myKey' | 'myOtherKey', string>) {
    emit('update:value', value)
  }
})

// if you want to bind props.value.myKey you'd need to wrap it in another get/set
// if you have multiple properties you have to duplicate this for each property
const myKey = computed({
  get() {
    return props.value.myKey
  },
  set(myKey: string) {
    internalValue.value = { 
      ...internalValue.value,
      myKey 
    }
  }
})

// using the usePatchRef composition abstracts the additional get/set for each property
const myKey = usePatchRef(internalValue, 'myKey')
const myOtherKey = usePatchRef(internalValue, 'myOtherKey')
```

## Arguments
| source | `Ref<T>` |
| key    | `keyof T` |

## Returns
`Ref<keyof T>`


stackoverfloweth (approved):
seems straight forward, don't forget to add it to the main readme as well

src/usePatchRef/usePatchRef.ts
```diff
@@ -0,0 +1,15 @@
+import { Ref, computed } from 'vue'
+
+export function usePatchRef<T extends Record<string, unknown>, P extends keyof T>(source: Ref<T>, key: P): Ref<T[P]> {
```
znicholasbrown, line 3: Could you add a docstring to this? Might make it easier to understand when someone is reading a component

PR #139, opened 2022-12-08
Feature: useValidation and useValidationObserver compositions

pleek91:
# Description
Observation only validation that supports nesting
- [useValidation](https://github.com/PrefectHQ/vue-compositions/tree/use-validation/src/useValidation)
- [useValidationObserver](https://github.com/PrefectHQ/vue-compositions/tree/use-validation/src/useValidationObserver)


src/useValidation/ValidationExecutor.ts
```diff
@@ -0,0 +1,37 @@
+import { ValidationRule } from './useValidation'
+import { ValidationAbortedError } from './ValidationAbortedError'
+import { NoInfer } from '@/types/generics'
+
+export class ValidationRuleExecutor<T> {
+  private controller = new AbortController()
```
znicholasbrown, line 6: TIL about `AbortController`, really nice use of that.

src/useValidation/useValidation.ts
```diff
@@ -0,0 +1,107 @@
+import { ComponentInternalInstance, computed, ComputedRef, getCurrentInstance, inject, onMounted, onUnmounted, ref, Ref, watch } from 'vue'
+import { ValidationAbortedError } from './ValidationAbortedError'
+import { ValidationRuleExecutor } from './ValidationExecutor'
+import { NoInfer } from '@/types/generics'
+import { MaybePromise, MaybeRef } from '@/types/maybe'
+import { isUseValidationObserver, UseValidationObserver, ValidationObserverUnregister, VALIDATION_OBSERVER_INJECTION_KEY } from '@/useValidationObserver/useValidationObserver'
+import { getSymbolForInjectionKey } from '@/utilities/symbols'
+
+const USE_VALIDATION_SYMBOL: unique symbol = Symbol('UseValidationSymbol')
```
znicholasbrown, line 9: This symbol is used only for the validation typeguard, yeah? Is there another use I'm missing?

pleek91: Correct. This is a technique I stole from vue's source code. By creating a `unique symbol` and not exporting it you can check for it within this project but it will never show up in the types when you use this composition. Which produced a much simpler and much safer type guard rather than checking for the presence of specific keys.

src/useValidationObserver/README.md
```diff
@@ -0,0 +1,33 @@
+# useValidationObserver
+The `useValidationObserver` combines
```
stackoverfloweth, line 2: ... suspense

src/useValidationObserver/useValidationObserver.ts
```diff
@@ -0,0 +1,116 @@
...
+    for (const key of keys) {
+      const validation = validations[key]
+
+      if (validation.pending) {
+        return true
+      }
+
+    }
+
+    return false
```
stackoverfloweth, line 89:
```suggestion
    return keys.some(key => validations[key].pending)
```

src/useValidationObserver/useValidationObserver.ts
```diff
@@ -0,0 +1,116 @@
...
+
+    keys.forEach(key => {
+      const validation = validations[key]
+
+      if (validation.error) {
+        errors.push(validation.error)
+      }
+    })
+
+    return errors
```
stackoverfloweth, line 74:
```suggestion
    return keys
      .map(key => validations[key].error)
      .filter(error => !!error)
```

stackoverfloweth (approved):
death to vee-validate! 🔪

src/useValidation/useValidation.ts
pleek91: Following a conversation with @stackoverfloweth these guards and symbols ended up being removed. We realized ultimately the guards were not needed.

znicholasbrown: Way ahead of me, thanks for the update 👍🏻

PR #149, opened 2022-12-16
Feature: added state to useValidation

stackoverfloweth:


src/useValidation/useValidation.ts
```diff
@@ -29,6 +31,10 @@ export function useValidation<T>(
   const valid = computed(() => error.value === '')
   const invalid = computed(() => !valid.value)
   const pending = ref(false)
+  const initialValue = unref(value)
```
pleek91, line 34: What if it's an object or an array? 

Also validation doesn't happen unless mounted. But the value can change before then. So this won't match expectation in all cases

src/useValidation/useValidation.ts
```diff
@@ -45,16 +51,27 @@ export function useValidation<T>(
     }
 
     pending.value = false
+    validated.value = true
```
pleek91, line 54: I think this has value.

src/useValidation/useValidation.ts
```diff
@@ -29,6 +31,10 @@ export function useValidation<T>(
   const valid = computed(() => error.value === '')
   const invalid = computed(() => !valid.value)
   const pending = ref(false)
+  const initialValue = unref(value)
+  const touched = ref(false)
+  const dirty = ref(false)
```
pleek91, line 36: These can be easily determined by the user at implementation. And the composition doesn't use them internally. So I don't think these offer enough value.

stackoverfloweth:
@pleek91 I only kept what you said had value and fixed some eslint warnings. I will make a follow-up change to prefect-design to replace it's `State` model with something more a kin to the type `UseValidation`

pleek91 (approved):

PR #150, opened 2022-12-19
Feature: moved simplified validation state from prefect-design

stackoverfloweth:
rather than moving in the route of [simplifying prefect-design](https://github.com/PrefectHQ/prefect-design/pull/556) I decided to move the model to `useValidation`.

since prefect-design has vue-compositions as a dependency, it makes sense that prefect-design gets updated to use this model. This enables the best devex

```ts
const { error, state } = useValidation(accountId, 'Account')
```

vs if prefect-design owns the model we need

```ts
const { error, valid, pending, validated } = useValidation(accountId, 'Account')
const state = reactive({ valid, pending, validated })
```

### DIFF
--- a/src/useValidation/useValidation.ts
+++ b/src/useValidation/useValidation.ts
@@ -1,20 +1,22 @@
-import { computed, ComputedRef, onMounted, onUnmounted, reactive, ref, Ref, watch } from 'vue'
+import { computed, onMounted, onUnmounted, reactive, ref, ToRefs, watch } from 'vue'
 import { NoInfer } from '@/types/generics'
 import { MaybePromise, MaybeRef } from '@/types/maybe'
 import { ValidationAbortedError } from '@/useValidation/ValidationAbortedError'
 import { ValidationRuleExecutor } from '@/useValidation/ValidationExecutor'
-import { ValidationState } from '@/useValidation/validationState'
 import { ValidationObserverUnregister, VALIDATION_OBSERVER_INJECTION_KEY } from '@/useValidationObserver/useValidationObserver'
 import { injectFromSelfOrAncestor } from '@/utilities/injection'
 
-export type UseValidation = {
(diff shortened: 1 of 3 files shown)


src/useValidation/validationState.ts
```diff
@@ -0,0 +1,5 @@
+export type ValidationState = {
```
pleek91, line 1: Something I've been doing recently is including a reactive of all the refs a composition returns. This is pretty close to that but without a few of the refs. Its a nice devx to be able to work with a reactive vs refs when able and I think it makes sense for this composition to expose everything both as a ref and in this reactive state. 

prefect-design can just use the pieces it needs (I'd have prefect design's `state` still just be the limited set that it actually uses).

stackoverfloweth: can you give me an example of where you do this? I'm curious how to define it's type and what to name the property

pleek91: That's what I would do

src/useValidation/useValidation.ts
```diff
@@ -52,13 +56,22 @@ export function useValidation<T>(
     return valid.value
   }
 
-  const validation: UseValidation = {
+  const state = reactive({
```
pleek91, line 59: One last thought. Would it make more sense for this reactive object to be named `validation`? `state` matches up nicely with prefect-design but it make less sense here as a standalone thing. I don't feel strongly, approving this either way. 

```typescript
const { validation } = useValidation()
```

pleek91 (approved):
Good enhancement 👍

src/useValidation/useValidation.ts
stackoverfloweth: I think I like `state` more, even without considering it's alignment with how we currently use vee-validate in prefect-design.

PR #156, opened 2022-12-19
Feature: useRouteQuery

pleek91:
# useRouteQuery
The `useRouteQuery` composition is a reactive wrapper around the query property from vue-router's `useRoute` composition. It's primary purpose is to batch route updates so that updating multiple query params at once is possible without having to set the whole query at once. 

## Example
```typescript
import { useRouteQuery } from '@prefecthq/vue-compositions'

const { query, clear, set, remove } = useRouteQuery()

query.foo = 'foo'
query.bar = 'bar'
query.baz = 'baz'

set('foo', 'foo-updated')

remove('baz')

// all of the above operations will only produce one route change with the route
// ?foo=foo-updated&bar=bar
```

## Arguments
None

## Returns
```typescript
type UseRouteQuery = {
  query: LocationQuery,
  clear: () => void,
  set: (key: string, value: string) => void,
  remove: (key: string) => void,
}
```


src/useRouteQuery/useRouteQuery.ts
```diff
@@ -0,0 +1,112 @@
...
+type QueryOperation = (query: LocationQuery) => LocationQuery
+
+export function useRouteQuery(): UseRouteQuery {
+  const route = useRoute()
+  const router = useRouter()
+  const operations = reactive<QueryOperation[]>([])
+  let interval: ReturnType<typeof setTimeout>
+
+  const clear = (): void => {
+    router.push({ query: {} })
```
stackoverfloweth, line 21: are you worried at all about conflicts between 2 components that both want to use this composition? I wonder if it would be worth the complexity to have `clear` only remove values that were explicitly added by this instance. I don't feel strongly, just thinking out loud

src/useRouteQuery/useRouteQuery.ts
```diff
@@ -0,0 +1,112 @@
...
+  let query: LocationQuery = { ...currentQuery }
+
+  for (const operation of operations) {
+    query = operation(query)
+  }
+
+  return query
+}
+
+function updateOperation(key: string, value: string): QueryOperation {
```
stackoverfloweth, line 96: nit, this function name isn't super clear to me. As is, I think it's more like `getUpdateOperation`, but if it were me, I might also take the `operations` as a param and call it `queueUpdateOperation`, feels like it works nicely to `queue` operations, and then eventually `apply` operations?

stackoverfloweth (approved):
small questions, looks great as is

src/useRouteQuery/useRouteQuery.ts
pleek91: No not really. This composition is designed to give you the whole query. I think it would be more confusing if it tried to scope the query just to specific properties.

src/useRouteQuery/useRouteQuery.ts
pleek91: I don't want to over think this considering these are private methods that are not exported. I updated the naming of the operation methods to be `updateOperationFactory` and `deleteOperationFactory` which accurately describes what they do. I also broke out the creation of an operation from the `operations.push` which makes it more readable IMO.

PR #159, opened 2022-12-20
Enhancement: subscription promise types

pleek91:
# Description
Fixes the types returned by the `promise` method on a subscription so that the response is not possibly undefined. 

```typescript
function action(): boolean {}

const { response } = await useSubscription(action).promise()

// before
response // boolean | undefined

// after
response // boolean

### CODE REVIEW

File: src/useSubscription/types/subscription.ts [hunk 1]
```diff
@@ -1,3 +1,4 @@
+import { Ref } from 'vue'
```
stackoverfloweth, line 1: seems `Ref` is unused 
```suggestion
```

stackoverfloweth (approved):

PR #164, opened 2023-01-02
Feature: Enhance useRouteQueryParam and add useRouteQueryParams composition

pleek91:
# Description
Two compositions were updated and one was added

## `useRouteQuery`
This composition didn't change functionally. But its been refactored to be a singleton. This further enhances the compositions ability to batch updates to the route query. Now any calls to `useRouteQueryParam` or `useRouteQueryParams` will batch. That way which composition is used doesn't change how updates are batched and the developer can use the composition that makes sense for the implementation. 

## `useRouteQueryParam`
Rewrote the existing `useRouteQueryParam` to allow for type safe formatting of values and to use the `useRouteQuery` composition. Should be completely backwards compatible with existing use cases

https://github.com/PrefectHQ/vue-compositions/tree/main/src/useRouteQueryParam

## `useRouteQueryParams`
A new composition for binding to multiple query params at once. Supports nested and optional params. 

https://github.com/PrefectHQ/vue-compositions/tree/use-route-query-params/src/useRouteQueryParams


package.json
```diff
@@ -66,6 +67,7 @@
     }
   },
   "dependencies": {
+    "flat": "^5.0.2",
```
pleek91, line 70: This package isn't used anymore. I'll remove this.

src/useRouteQuery/useRouteQuery.ts
```diff
@@ -1,90 +1,111 @@
...
-
-      return true
-    },
-  })
-
   const update = (): void => {
     const query = applyQueryOperations(route.query, operations)
 
+    operations.splice(0)
+
```
pleek91, line 24: This was a bugfix. Clear the operations array when the route is updated.

src/useRouteQuery/useRouteQuery.ts
```diff
@@ -1,90 +1,111 @@
...
     router.push({ query })
   }
 
   watch(operations, () => {
+    if (!operations.length) {
+      return
+    }
```
pleek91, line 31: This is a bugfix. No need to update the route if there are no operations.

src/useRouteQueryParam/formats/RouteParam.ts
```diff
@@ -0,0 +1,105 @@
...
+// adding any here so RouteParamClass can be used without passing a generic
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export type RouteParamClass<T = any> = new (args: RouteParamClassArgs<T>) => RouteParam<T>
+
+export function isRouteParamClass(value: unknown): value is RouteParamClass<unknown> {
+  return typeof value === 'function' && 'IS_ROUTE_PARAM' in value && value.IS_ROUTE_PARAM == IS_ROUTE_PARAM_SYMBOL
+}
+
+export abstract class RouteParam<T> {
+  public static IS_ROUTE_PARAM = IS_ROUTE_PARAM_SYMBOL
```
pleek91, line 22: I think this is incorrect. Needs to use the symbol itself as the key.

src/useRouteQueryParam/formats/RouteParam.ts
```diff
@@ -0,0 +1,105 @@
...
+const IS_ROUTE_PARAM_SYMBOL: unique symbol = Symbol()
+
+export type RouteParamClassArgs<T> = {
+  key: string,
+  defaultValue: T | T[],
+}
+
+// adding any here so RouteParamClass can be used without passing a generic
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export type RouteParamClass<T = any> = new (args: RouteParamClassArgs<T>) => RouteParam<T>
```
stackoverfloweth, line 15: do you know why we can't use `InstanceType<RouteParam<T>>`?

src/useRouteQueryParam/formats/RouteParam.ts
```diff
@@ -0,0 +1,105 @@
...
+    if (value === undefined) {
+      return routeQuery.remove(this.key)
+    }
+
+    const values = asArray(value)
+    const strings = values.map(value => this.safeFormatValue(value)).filter(isNotInvalidRouteParamValue)
+
+    if (this.multiple) {
+      this.defaultValue = []
+    }
```
stackoverfloweth, line 66: why update the defaultValue here?

src/useRouteQuery/useRouteQuery.ts
```diff
@@ -1,90 +1,111 @@
...
+
+    const get = (key: string): LocationQueryValue | LocationQueryValue[] => {
+      return query[key]
+    }
+
+    const remove = (key: string): void => {
+      delete query[key]
+    }
+
+    const query: LocationQuery = new Proxy({}, {
```
stackoverfloweth, line 57: what's the benefit of using a proxy here? couldn't you just as easily define some non-exported methods for `get`/`set`/`delete`?

src/useRouteQuery/useRouteQuery.ts
```diff
@@ -1,90 +1,111 @@
...
+
+        return true
+      },
+      ownKeys() {
+        const query = applyQueryOperations(route.query, operations)
+
+        return Object.keys(query)
+      },
+      has(target, property) {
+        return property in route.query
```
stackoverfloweth, line 95: why did you choose not to call `applyQueryOperations` here like you do for other `get`-ish functions?

src/useRouteQueryParam/formats/RouteParam.ts
pleek91: Fixed. Included a test for this as well.

src/types/maybe.ts
```diff
@@ -1,6 +1,7 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
 import { Ref, UnwrapRef } from 'vue'
 
-export type MaybePromise<T> = T | Promise<T>
-export type MaybeRef<T> = T | Ref<T>
-export type MaybeUnwrapRef<T> = T | UnwrapRef<T>
-export type MaybeArray<T> = T | T[]
\ No newline at end of file
+export type MaybePromise<T = any> = T | Promise<T>
```
znicholasbrown, line 4: Is `unknown` incorrect here?

src/useRouteQueryParam/useRouteQueryParam.ts
```diff
@@ -1,55 +1,46 @@
...
+
+function isDefaultValue<T>(value: T | RouteParamClass): value is T {
+  return !isRouteParamClass(value)
+}
 
-  const unwatchRoute = watch(route, (newRoute, oldRoute) => {
-    const newValue = getRouteQueryParam(newRoute, param) ?? defaultValue
-    const oldValue = getRouteQueryParam(oldRoute, param) ?? defaultValue
-    const matched = matchValueType(oldValue, newValue)
+export function useRouteQueryParam(key: string, defaultValue?: string): Ref<string>
```
stackoverfloweth, line 13: I know this composition has gone through a few iterations, I just want to say that the way it turned out is quite easy to understand and just generally fantastic

src/useRouteQueryParams/useRouteQueryParams.ts
```diff
@@ -0,0 +1,65 @@
+import { Ref } from 'vue'
+import { NoInfer } from '@/types/generics'
+import { isRouteParamClass, RouteParamClass } from '@/useRouteQueryParam/formats'
+import { useRouteQueryParam } from '@/useRouteQueryParam/useRouteQueryParam'
+
+type AnyRecord = Record<string, unknown>
```
stackoverfloweth, line 6: probably more like `UnknownRecord`, or `UnknownObject`. We really should export types like this somewhere

src/types/maybe.ts
pleek91: Great question. I wanted to be able to use these types without adding a generic every time. And in `MaybePromise<T>` `T` is unknown but it requires you pass it in because there is no default. Which is why I added the `any`. But taking another look at it I'm realizing I can set the default to `unknown` which seems much safer. 

Updated!

src/useRouteQueryParams/useRouteQueryParams.ts
```diff
@@ -0,0 +1,65 @@
...
+    const property = schema[key]
+    const propertyDefault = defaultValue[key]
+
+    if (isRouteParamClass(property)) {
+      params[key] = useRouteQueryParam(prefixed(key), property, propertyDefault)
+
+      return params
+    }
+
+    if (isRouteParamSchema(property)) {
```
stackoverfloweth, line 53: I'm guessing you did this to be highly explicit, but it seems you could replace this line with `else`, since this type guard just asserts that it's not `isRouteParamClass`

src/useRouteQuery/useRouteQuery.ts
znicholasbrown: Looking at this questions elsewhere and it seems to be, oof. Ignore me!

pleek91: Reassignment won't work on anything that is `reactive`. I think setting the length would work though I don't personally see a benefit over `splice`. But if there's a reason that's better I'll test it and if it works we can do that. 

I went with `splice(0)` purely because that was one of the solutions I found in a vue github issue.

stackoverfloweth (approved):
Might be the nicest composition in this library. Great job as always @pleek91 👏

src/useRouteQuery/useRouteQuery.ts
```diff
@@ -1,90 +1,111 @@
...
-    set,
-    get,
-    remove,
-    query,
+  const useRouteQuery = (): UseRouteQuery => {
+    router = useRouter()
+    route = useRoute()
+
+    const clear = (): void => {
+      router.push({ query: {} })
```
znicholasbrown, line 42: I'm surprised to see this method pushing to the router instead of resetting the `query` variable and relying on its  setter

src/useRouteQueryParam/formats/RouteParam.spec.ts
```diff
@@ -0,0 +1,11 @@
+import { describe, expect, it } from 'vitest'
+import { RouteParam, isRouteParamClass } from '@/useRouteQueryParam/formats/RouteParam'
+
+describe('RouteParam', () => {
+
+  it('type guard is valid', () => {
+    expect(isRouteParamClass(RouteParam)).toBe(true)
```
stackoverfloweth, line 7: if this test were to fail, wouldn't the build itself fail?

src/useRouteQuery/useRouteQuery.ts
znicholasbrown: Yeah this is probably just on me; I've always hated that `splice` operates on the array in place because it enables operations like this which don't feel entirely clear.

znicholasbrown (approved):
LGTM

src/useRouteQueryParam/formats/RouteParam.spec.ts
pleek91: No, this is essentially testing the logic inside the `isRouteParamClass` function. That could be totally jacked up and the build would still pass as long as the logic returned a boolean.

src/useRouteQuery/useRouteQuery.ts
pleek91: hmmm, I guess I could do that. But the `query` variable is just a proxy wrapper around `route.query` and the `operations` queue. And objects don't have a specific way of emptying them so it wasn't something I thought to support on the `query` proxy itself. 

Updating the url directly seems like the simplest thing to do here IMO.

src/useRouteQueryParams/useRouteQueryParams.ts
pleek91: If it wasn't for typescript I wouldn't need this if (or an else) at all. But anytime `Object.keys` is involved typescript forces you to be super explicit. I did try using an `else` but typescript doesn't really do well with type guards and else. It doesn't get the correct type and then I have a type error in the `getSchemaRouteQueryParams` call. 

Open to suggestions here. This was the only way I could figure out to appease typescript.

src/useRouteQueryParams/useRouteQueryParams.ts
pleek91: Yeah this is a dumb type so I could be lazy. Its really not even that accurate. `Record<string | number | symbol, unknown>` would be a better `AnyRecord` type. But I don't want to check for `number` or `symbol` keys.

src/useRouteQueryParam/useRouteQueryParam.ts
pleek91: ❤️

src/useRouteQuery/useRouteQuery.ts
znicholasbrown: Yeah fair enough 🤷🏻

src/useRouteQuery/useRouteQuery.ts
pleek91: Wasn't intentional. And using `applyOperations` here makes sense because you should be able to check for the presence of a key even if the operations haven't been applied. Good catch!

src/useRouteQuery/useRouteQuery.ts
pleek91: I'm not totally following the methods piece but the idea here is that the `query` const can be used exactly the same as the `route.query` object. Anywhere you use `route.query` in code you could be able to use `query`.

src/useRouteQueryParam/formats/RouteParam.ts
pleek91: I tried a few different things here and this was what I could get to work. But `InstanceType` is just that. Its the instance and we want a constructor here. 

Might not be what you're asking. So leaving this unresolved.

src/types/maybe.ts
```diff
@@ -1,6 +1,7 @@
...
+export type MaybeRef<T = any> = T | Ref<T>
+export type MaybeUnwrapRef<T = any> = T | UnwrapRef<T>
+export type MaybeArray<T = any> = T | T[]
```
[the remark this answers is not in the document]
pleek91: These got updated to `T = unknown` thanks to @znicholasbrown 's comment.

src/useRouteQueryParams/useRouteQueryParams.ts
stackoverfloweth: hmmm bummer, the only alternative I can think of is casting the `property` on line 44. Don't think that's much better

src/useRouteQueryParam/formats/RouteParam.ts
pleek91: I have no idea.... I feel like this was left over from a previous iteration. I'll look into it but I think this should just be removed.

pleek91: Ohh I remember. This was from a previous iteration but the thought is still valid. The thought being, if there is a default value for an array param. That default is used if the param does not exist. But that also means there's no way to get to an empty array. 

I haven't rethought about that problem since the last refactor. So taking a chance to think about it again. What are your thoughts on that?

src/useRouteQueryParams/useRouteQueryParams.ts
pleek91: The thing that always gets me is typescript is technically correct. There could be other properties at run time (even though there aren't likely to be). 

Since this works and is pretty readable I'm going to leave it as is for now.

src/useRouteQueryParam/formats/RouteParam.ts
stackoverfloweth: if that is in case needed, we should squirrel this away in a function that explains what we're up to

src/useRouteQueryParam/formats/RouteParam.ts
stackoverfloweth: Maybe idk typescript but I thought `function(input:SomeClass)` is a function that takes an instance of `SomeClass` as it's only argument. I would expect `InstanceType` to get me the underlying type for this instance. But it doesn't work here, so clearly I'm mistaken

src/types/maybe.ts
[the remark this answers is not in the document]
stackoverfloweth: still 🧠, maybe even 🧠 🧠

src/useRouteQueryParam/formats/RouteParam.ts
pleek91: You're explaining the types correctly. But what you're missing is the functions in this PR don't accept a `RouteParam` (instance) but a `RouteParamClass` (constructor). 

image

stackoverfloweth: right, my original comment was that `RouteParamClass` seemed like an unnecessary abstraction. I expected you to be able to use `InstaceType<RouteParam<T>>`

pleek91: Maybe the definition for `InstanceType` will help explain why its not what you think it is 

```typescript
type InstanceType<T extends abstract new (...args: any) => any> = T extends abstract new (...args: any) => infer R ? R : any;
```

pleek91: As far as I know (or have been able to find) there isn't a way to go from an instance (RouteParam) to its constructor type.

src/useRouteQueryParam/formats/RouteParam.ts
pleek91: This needs some more work. But for now I'm going to remove these lines and we can support the use case like this
```typescript
const states = useRouteQueryParam('state', [])

if(states.length === 0) {
  states.value = ['running'] // this is the "default"
}
```

PR #170, opened 2023-01-06
BugFix: Fix array query params not updating and fix query params flashing 

pleek91:
# Description
Fixes two bugs. Also has the proxy throw an error if an invalid property is used rather than just silently failing.

### DIFF
--- a/src/useRouteQuery/useRouteQuery.ts
+++ b/src/useRouteQuery/useRouteQuery.ts
@@ -65,7 +65,7 @@ function factory(): () => UseRouteQuery {
         return query[property]
       },
       set(target, property, value) {
-        if (typeof property === 'symbol' || typeof value !== 'string') {
+        if (typeof property === 'symbol') {
           return Reflect.set(target, property, value)
         }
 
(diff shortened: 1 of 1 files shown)


src/useRouteQuery/useRouteQuery.ts
```diff
@@ -17,12 +23,12 @@ function factory(): () => UseRouteQuery {
...
 
-  const update = (): void => {
+  const update = async (): Promise<void> => {
     const query = applyQueryOperations(route.query, operations)
 
-    operations.splice(0)
+    await router.push({ query })
 
-    router.push({ query })
+    operations.splice(0)
```
pleek91, line 31: This fixes the flashing issue

src/useRouteQuery/useRouteQuery.ts
```diff
@@ -54,19 +60,19 @@ function factory(): () => UseRouteQuery {
...
-          return Reflect.get(target, property)
+          throw new InvalidRouteQueryProperty()
         }
 
         const query = applyQueryOperations(route.query, operations)
 
         return query[property]
       },
       set(target, property, value) {
-        if (typeof property === 'symbol' || typeof value !== 'string') {
```
pleek91, line 68: This fixes arrays not being set.

stackoverfloweth (approved):

PR #171, opened 2023-01-06
Enhancement: Better support null and undefined values in useRouteQueryParam

pleek91:
# Description
Adds default support for undefined params and adds nullable versions of all the formatters. 

I also added a date formatter since they're wasn't one already.


src/useRouteQueryParam/formats/NullableBooleanRouteParam.ts
```diff
@@ -0,0 +1,27 @@
+import { LocationQueryValue } from 'vue-router'
+import { InvalidRouteParamValue } from '@/useRouteQueryParam/formats/InvalidRouteParamValue'
+import { RouteParam } from '@/useRouteQueryParam/formats/RouteParam'
+
+export class NullableBooleanRouteParam extends RouteParam<boolean | null> {
+
+  protected override parse(value: LocationQueryValue): boolean | null {
```
stackoverfloweth, line 7: how does this ever return `null` ?

src/useRouteQueryParam/formats/NullableBooleanRouteParam.ts
```diff
@@ -0,0 +1,27 @@
...
+    if (value === 'true') {
+      return true
+    }
+
+    if (value === 'false') {
+      return false
+    }
```
stackoverfloweth, line 14: any reasonable way to reuse the `BooleanRouteParam` class here?

stackoverfloweth (approved):

src/useRouteQueryParam/formats/NullableBooleanRouteParam.ts
pleek91: Not without refactoring something. `BooleanRouteParam` extends `RouteParam<boolean>` which is what sets the types for the `parse` and `format` methods. Extending `BooleanRouteParam` and returning `null` throws ts errors. 

Very open to ideas here. But didn't think it was worth putting to much effort into reusing a few lines of code.

src/useRouteQueryParam/formats/NullableBooleanRouteParam.ts
pleek91: id didn't. fixed! Good catch.

PR #179, opened 2023-01-13
Enhancement: Expand functionality of validation rules

pleek91:
# Description
Convert the third argument of a ValidationRule from a single `signal` to a context argument that exposes additional information about the context of the validation. 

```typescript
export type ValidationRuleContext<T> = {
  signal: AbortSignal,
  source: string | undefined,
  previousValue: T | undefined,
}
```

This allows rules to be more dynamic based on what triggered the validation. 

Also added `void` to the return type for a `ValidationRule`. If a rule returns void its a no op. Meaning it will remain in whatever state it was previously.


stackoverfloweth (approved):
very nice, feels like this really justifies the extra complexity ❤️

PR #180, opened 2023-01-14
BugFix: useValidation async validation enhancments

pleek91:
# Description
Fixes some logic errors around async validation, skipped rules, and aborting.

### DIFF
--- a/src/useValidation/ValidationExecutor.ts
+++ b/src/useValidation/ValidationExecutor.ts
@@ -1,6 +1,5 @@
 import { ValidationRule } from '@/useValidation/useValidation'
 import { ValidationAbortedError } from '@/useValidation/ValidationAbortedError'
-import { ValidationSkippedError } from '@/useValidation/ValidationSkippedError'
 
 type Validate<T> = {
   value: T,
@@ -34,9 +33,8 @@ export class ValidationRuleExecutor<T> {
         throw new ValidationAbortedError()
       }
 
-      // we cannot stop looping when a rule skips...
-      if (result === undefined) {
-        throw new ValidationSkippedError()
+      if (skipped(result) || passed(result)) {
+        continue
       }
 
       if (result === false) {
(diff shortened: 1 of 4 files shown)


src/useValidation/ValidationExecutor.ts
```diff
@@ -18,23 +18,23 @@ export class ValidationRuleExecutor<T> {
...
         source,
         signal,
         previousValue,
       })
 
-      if (result === undefined) {
-        return
+      if (signal.aborted) {
+        throw new ValidationAbortedError()
+      }
```
pleek91, line 34: Moved this to after the rule execution so that we can abort after each rule. Otherwise if validation is aborted during an async rule the rule will still return an error.

src/useValidation/ValidationExecutor.ts
```diff
@@ -18,23 +18,23 @@ export class ValidationRuleExecutor<T> {
...
+
+      if (skipped(result) || passed(result)) {
+        continue
```
pleek91, line 37: Makes sure we don't return an error for these cases.

src/useValidation/useValidation.ts
```diff
@@ -33,7 +33,7 @@ export type ValidationRuleContext<T> = {
   previousValue: T | undefined,
 }
 
-export type ValidationRule<T> = (value: T, name: string, meta: ValidationRuleContext<T>) => MaybePromise<boolean | string | void>
+export type ValidationRule<T> = (value: T, name: string, meta: ValidationRuleContext<T>) => MaybePromise<boolean | string | undefined>
```
pleek91, line 36: using `undefined` rather than `void` which is more accurate to whats happening.

src/useValidation/useValidation.ts
```diff
@@ -83,21 +83,18 @@ export function useValidation<T>(
...
       })
 
-      if (result === undefined) {
-        return valid.value
-      }
-
       error.value = result
+      pending.value = false
+      validated.value = true
+      previousValueRef.value = valueRef.value
```
pleek91, line 89: Makes sure these only get set if validation is not aborted. Otherwise it can appear that validation is complete sooner than it actually is.

stackoverfloweth (approved):
Appreciate the comments @pleek91

PR #194, opened 2023-02-02
Enhancement: Sync value to input ref if debounced ref is set in useDebouncedRef

pleek91:
# Description
`useDebouncedRef` returns `Ref<T>` but if you set the value of that ref nothing happens. The intent of this composition is that the refs are the same but that there is a reactivity delay for the debounced version. This update will sync the value from the debounced ref back to the input ref (immediately). 

I also added some unit tests for this composition

### DIFF
--- a/src/useDebouncedRef/useDebouncedRef.spec.ts
+++ b/src/useDebouncedRef/useDebouncedRef.spec.ts
@@ -21,6 +21,8 @@ describe('useDebouncedRef', () => {
     const output = useDebouncedRef(input, 100)
     input.value = 2
 
+    expect(output.value).toBe(1)
+
     await vi.runAllTimersAsync()
 
     expect(output.value).toBe(2)
--- a/src/useDebouncedRef/useDebouncedRef.spec.ts
+++ b/src/useDebouncedRef/useDebouncedRef.spec.ts
@@ -1,0 +1,38 @@
+import { afterEach, vi, expect, describe, it } from 'vitest'
+import { ref } from 'vue'
+import { useDebouncedRef } from '@/useDebouncedRef/useDebouncedRef'
+
+afterEach(() => {
+  vi.useRealTimers()
+})
+
+describe('useDebouncedRef', () => {
+
+  it('it syncs the initial value', () => {
+    const input = ref(1)
+    const output = useDebouncedRef(input, 100)
+
+    expect(output.value).toBe(1)
+  })
+
+  it('it debounces updates to the value', async () => {
+    vi.useFakeTimers()
+    const input = ref(1)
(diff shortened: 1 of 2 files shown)


src/useDebouncedRef/useDebouncedRef.spec.ts
```diff
@@ -0,0 +1,38 @@
...
+    expect(output.value).toBe(2)
+  })
+
+  it('it syncs the debounced value back to the input', () => {
+    vi.useFakeTimers()
+    const input = ref(1)
+    const output = useDebouncedRef(input, 100)
+    output.value = 2
+
+    expect(input.value).toBe(2)
```
stackoverfloweth, line 37: how does this pass without `runAllTimers`?

src/useDebouncedRef/useDebouncedRef.spec.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  it('it debounces updates to the value', async () => {
+    vi.useFakeTimers()
+    const input = ref(1)
+    const output = useDebouncedRef(input, 100)
+    input.value = 2
+
+    await vi.runAllTimersAsync()
+
+    expect(output.value).toBe(2)
```
stackoverfloweth, line 28: should there be some parallel test that verifies that the value is NOT changed until after the timers run?

stackoverfloweth (approved):
excellent use case for tests! Well done @pleek91

src/useDebouncedRef/useDebouncedRef.spec.ts
pleek91: There's no debounce syncing from output to input. Only when setting input will there be a debounce in syncing that value to the output.

stackoverfloweth: ah, yeah I had that backwards in my mind.

stackoverfloweth: I wonder if the contract I envisioned is actually more straightforward?

stackoverfloweth:
```
const myActualValue = ref(1)
const theDebouncedVersion = useDeboundedRef(myActualValue)

myActualValue.value = 2 // happens immediately
theDebouncedVersion.value = 2 // awaits the delay
```

pleek91: oh interesting. I think of the debounced ref as a copy that is synced using a debounce. The lines get a little blurred though when you can set either value. I could see it both ways. But since we're already using it the input -> debounce -> output way I think we should keep it as is.

PR #216, opened 2023-03-20
Feature: useNow composition

pleek91:
# useNow
The `useNow` composition can be used to get a reactive current date instance.

## Example
```typescript
import { useNow } from '@prefecthq/vue-compositions'

const { now } = useNow(element)
```

## Arguments
| Name      | Type         | Default            |
|-----------|--------------|--------------------|
| immediate | `boolean`    | `true`             |

## Returns
| Name   | Type         |
|--------|--------------|
| now    | 'Date'       |
| pause  | `() => void` |
| resume | `() => void` |


znicholasbrown:
Fair enough @pleek91 👍🏻

PR #226, opened 2023-04-06
Enhancement: `useSubscription` reactive options [RFC]

znicholasbrown:
This PR slightly changes the behavior of the `useSubscription` composition, which previously took a static `options` argument, to instead take either a static `options` argument or a `Ref`. The composition now reacts to options changes and modifies the resulting subscription accordingly. 

Example:
```ts
const num = ref(0)
const interval = ref(1000)

function incNum(): void {
  num.value += 1
}

function toggleInterval(): void {
  interval.value = interval.value === Infinity ? 1000 : Infinity
}

const subscriptionOptions = computed<SubscriptionOptions>(() => {
  return {
    interval: interval.value,
  }
})

useSubscription(incNum, [], subscriptionOptions)
```

In the above example, when `interval` is set to `Infinity`, `num` will stop incrementing. When `interval` is set to `1000` again, it will appropriately resume incrementing

### DIFF
--- a/src/useSubscription/types/subscription.ts
+++ b/src/useSubscription/types/subscription.ts
@@ -1,10 +1,11 @@
+import { MaybeRef } from '@/types/maybe'
 import Manager from '@/useSubscription/models/manager'
 import Subscription from '@/useSubscription/models/subscription'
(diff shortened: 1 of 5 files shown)


pleek91 (approved):
LGTM! Excited about this.

Issue #220, opened 2023-03-29
useIntersectionObserver doesn't handle reactive options


pleek91:
# Description
it would be nice if the composition would automatically handle reactive changes to things like rootMargin, threshold, etc.

PR #230, opened 2023-04-12
Enhancement: Update useVisibilityObserver to use reactive options

pleek91:
# Description
Updates the `options` argument to be `Maybe<UseIntersectionObserverOptions>` and updates the internal logic to tear down the existing observer and create a new one when options change. It does this while keeping track of elements that are currently being observed.

Closes https://github.com/PrefectHQ/vue-compositions/issues/220


src/useIntersectionObserver/useIntersectionObserver.ts
```diff
@@ -81,6 +94,11 @@ export function useIntersectionObserver(callback: UseIntersectionObserverCallbac
     disconnect()
   })
 
+  watch(optionsRef, () => {
+    console.log('here')
```
znicholasbrown, line 98:
```suggestion
```

znicholasbrown (approved):
Well this looks familiar

LGTM @pleek91, just 1 console log to remove

PR #231, opened 2023-04-13
Feature: added lifecycle option to useSubscription

stackoverfloweth:
when creating a subscription, you now have the option of passing in a `lifecycle` argument to override the default behavior to unsubscribe on component unmount. 

```ts
const livesForever = useSubscription(myCoolFunction, [], { lifecycle: 'app' })
```

even when the component that calls `useSubscription` in it's setup gets unmounted, the manager will retain the subscription and any data fetched for the life of the app.


pleek91 (approved):
Nice! I've wanted to add this for a while. Fun collaborating with you on this 👍

PR #233, opened 2023-04-19
Enhancement: Channel response reactivity

pleek91:
# Description
Each time a channel executes the action it sets the response value on the cannel itself and on each subscription to that channel. The response on a subscription is a `ref` which means assigning it triggers reactive effects. Trigger reactive effects can cascade and cause quite a bit of processing/rendering. 

This has caused issues with dependent subscriptions refreshing unnecessarily (fixed by `uniqueValueWatcher`) and computed values and component templates being re-evaluated. Technically this is correct because the ref was assigned and everything that depends on that ref should be re-evaluated. 

However this is usually not desired if the actual value of the subscription has not changed. If an action returns the same value as the previous execution, there is no need to trigger effects. 

Checking if the new action response is the same as the previous before assigning it fixes this issue. I was hesitant to introduce a deep equality check. But @znicholasbrown and I both believe the cost of this check is actually a net positive once considering all of the reactive effects that are no longer triggered.

### DIFF
--- a/src/useSubscription/models/channel.ts
+++ b/src/useSubscription/models/channel.ts
@@ -1,4 +1,5 @@
 /* eslint-disable max-classes-per-file */
+import { isEqual } from 'lodash'
 import Manager from '@/useSubscription/models/manager'
 import Subscription from '@/useSubscription/models/subscription'
 import {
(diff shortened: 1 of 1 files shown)


znicholasbrown (approved):
It should also be noted that the deep quality check that we're using comes from lodash and has extremely robust escape hatches so that the check is only as expensive as it needs to be.

PR #234, opened 2023-04-20
check is writable before reassigning in debounce

brandonreid:


pleek91 (approved):
For context we were seeing a warning from vue about not being able to write to the value because it was readonly. But we were not actually trying to write the debounced ref. It appeared that vue was tracking some nested refs inside the computed we were debouncing. Adding this should fix that warning and its a better pattern anyway.

PR #243, opened 2023-05-01
Enhancement: Remove nullish query param's keys

znicholasbrown:
This PR slightly changes the behavior of the `useRouteQuery` composition by deleting keys from the URL whose values have been set to a nullish value. 

This is nice because empty values clutter up the URL without much added value, e.g.:
```
https://some.url/?search=&tags=foo,bar
```
vs
```
https://some.url/?tags=foo,bar
```

### DIFF
--- a/src/useRouteQuery/useRouteQuery.ts
+++ b/src/useRouteQuery/useRouteQuery.ts
@@ -48,7 +48,11 @@ function factory(): () => UseRouteQuery {
     }
 
     const set = (key: string, value: LocationQueryValue | LocationQueryValue[]): void => {
-      query[key] = value
+      if (value?.length === 0) {
+        remove(key)
+      } else {
+        query[key] = value
+      }
     }
 
     const get = (key: string): LocationQueryValue | LocationQueryValue[] => {


src/useRouteQuery/useRouteQuery.ts
```diff
@@ -48,7 +48,11 @@ function factory(): () => UseRouteQuery {
     }
 
     const set = (key: string, value: LocationQueryValue | LocationQueryValue[]): void => {
-      query[key] = value
+      if (value?.length === 0) {
```
znicholasbrown, line 51: It might not be obvious from reading this in Github but `value`'s type ends up as `string | string[] | undefined`

pleek91:
So the issue here is `null` and empty strings are valid values for a query param. And `?foo` is how you express a null value for a param. And `?foo=` is how you express an empty string. And since query params can have default values. The presence of the param in the url is important to override the default. For example `useRouteQueryParam(foo, StringQueryParam, 'bar')` where `bar` is the default value. But the value could be changed to an empty string.

PR #258, opened 2023-05-22
Feature: `useBoolean` composition

znicholasbrown:
This PR introduces a `useBoolean` composition which is a handy utility wrapper for managing the state of a `ref<boolean>`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,7 @@ npm i --save @prefecthq/vue-compositions
 ```
 
 ## Compositions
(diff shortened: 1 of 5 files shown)


pleek91:
@znicholasbrown can you update the root README to include this composition. I also think the barrel needs to be updated.

znicholasbrown:
Good looks @pleek91, thank you - updated

pleek91 (approved):
Nice!

PR #276, opened 2023-07-05
BugFix: useSubscriptionWithDependencies can miss changes to args if a computed has nested reactive properties

pleek91:
# Description
Even though args is a `Ref` its still possible for a ref to have nested properties that are reactive objects. So deep watch is still needed here same as in `useSubscription`


znicholasbrown (approved):
thanks @pleek91

brandonreid (approved):

PR #293, opened 2023-08-16
Use event listener composition

stackoverfloweth:
# useEventListener

The `useEventListener` composition can be used to automatically handle setup and teardown of event listeners on the document or HTMElement scope. The options argument extends browser AddEventListenerOptions with `immediate` boolean, which defaults to `true` but when passed in as `false`, will prevent the composition from adding the listener automatically.

## Example

```typescript
import { useEventListener } from '@prefecthq/vue-compositions'

function handleEvent(event: MouseEvent) {
  // Respond to event
}
const element = ref<HTMLDivElement | undefined>()
useEventListener(element, 'keyup', handleEvent)
```

## Arguments

| Name      | Type                                                      | Default   |
|-----------|-----------------------------------------------------------|-----------|
| type      | `K (K extends keyof DocumentEventMap)`                    | None      |
| callback  | `(this: Document, event: DocumentEventMap[K]) => unknown` | None      |
| options   | `AddEventListenerOptions & { immediate: boolean }`        | None      |

## Returns

| Name   | Type        | Description                                       |
|--------|-------------|---------------------------------------------------|
| add    | () => void  | Manually attach the event listener (has no effect if the event listener already exists) |
| remove | () => void  | Manually detach the event listener                |

### DIFF
--- a/src/useEventListener/README.md
+++ b/src/useEventListener/README.md
@@ -1,0 +1,30 @@
+# useEventListener
+
+The `useEventListener` composition can be used to automatically handle setup and teardown of event listeners on the document or HTMElement scope. The options argument extends browser AddEventListenerOptions with `immediate` boolean, which defaults to `true` but when passed in as `false`, will prevent the composition from adding the listener automatically.
+
+## Example
+
+```typescript
+import { useEventListener } from '@prefecthq/vue-compositions'
+
+function handleEvent(event: MouseEvent) {
+  // Respond to event
+}
+const element = ref<HTMLDivElement | undefined>()
+useEventListener(element, 'keyup', handleEvent)
+```
+
+## Arguments
+
+| Name      | Type                                                      | Default   |
+|-----------|-----------------------------------------------------------|-----------|
+| type      | `K (K extends keyof DocumentEventMap)`                    | None      |
+| callback  | `(this: Document, event: DocumentEventMap[K]) => unknown` | None      |
+| options   | `AddEventListenerOptions & { immediate: boolean }`        | None      |
+
+## Returns
+
+| Name   | Type        | Description                                       |
+|--------|-------------|---------------------------------------------------|
+| add    | () => void  | Manually attach the event listener (has no effect if the event listener already exists) |
+| remove | () => void  | Manually detach the event listener                |
--- a/src/useEventListener/README.md
+++ b/src/useEventListener/README.md
@@ -2,6 +2,10 @@
 
 The `useEventListener` composition can be used to automatically handle setup and teardown of event listeners on the document or HTMElement scope. The options argument extends browser AddEventListenerOptions with `immediate` boolean, which defaults to `true` but when passed in as `false`, will prevent the composition from adding the listener automatically.
 
+The composition will return two methods `add` and `remove`. Calling add will trigger `addEventListener` on the target. Calling `remove` will trigger `removeEventListener` on the target.
+
+The composition uses a watcher to remove and re-add the eventListener automatically when the `target` changes. Note this will NOT execute if `options.immediate` is `false`, or if `remove` is called.
+
 ## Example
 
 ```typescript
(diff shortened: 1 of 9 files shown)


stackoverfloweth:
this PR also updates `useGlobalEventListener` to simply wrap new `useEventListener` with `document` as target.

src/useEventListener/README.md
```diff
@@ -0,0 +1,30 @@
...
+useEventListener(element, 'keyup', handleEvent)
+```
+
+## Arguments
+
+| Name      | Type                                                      | Default   |
+|-----------|-----------------------------------------------------------|-----------|
+| type      | `K (K extends keyof DocumentEventMap)`                    | None      |
+| callback  | `(this: Document, event: DocumentEventMap[K]) => unknown` | None      |
+| options   | `AddEventListenerOptions & { immediate: boolean }`        | None      |
```
pleek91, line 23: This has a default of `{ immediate: true }` right?

src/useEventListener/useEventListener.ts
```diff
@@ -0,0 +1,49 @@
...
+    toValue(target)?.addEventListener(key, callback, listenerOptions)
+  }
+
+  function remove(): void {
+    toValue(target)?.removeEventListener(key, callback, listenerOptions)
+  }
+
+  if (getCurrentScope()) {
+    onScopeDispose(() => remove())
+  }
```
pleek91, line 34: Could you use `tryOnScopeDispose` for this?

src/useEventListener/useEventListener.ts
```diff
@@ -0,0 +1,49 @@
...
+
+  if (immediate) {
+    add()
+  }
```
pleek91, line 38: So I believe if this composition is called in setup any template refs will be undefined. So rather than doing this in an onMount you're letting the watcher handle adding it when it becomes defined. Am I understanding that correctly?

src/useEventListener/useEventListener.ts
```diff
@@ -0,0 +1,49 @@
...
+
+  watch(() => toValue(target), () => {
+    remove()
+    add()
+  })
```
pleek91, line 43: I think this needs to be a little more sophisticated. If `immediate` is `false` I think this might add the listener anyway. Similar if `remove` is called but then the target changes this will automatically add the listener back.

pleek91 (commented):
Couple questions but overall this is sweet!

src/useEventListener/README.md
stackoverfloweth: yes, I'll update the readme accordingly

src/useEventListener/useEventListener.ts
stackoverfloweth: yes, but this will change when I address the comment below

src/useEventListener/useEventListener.ts
stackoverfloweth: I think I have a better, more elegant solution but it now seems to warrant some unit tests. Let me get some good tests written and I'll update this PR

src/useEventListener/useEventListener.spec.ts
```diff
@@ -0,0 +1,158 @@
...
+
+    // because reattaching would happen in watch
+    await timeout()
+
+    expect(originalAddEventListenerMock).toHaveBeenCalledOnce()
+    expect(updatedAddEventListenerMock).not.toHaveBeenCalled()
+  })
+
+})
+
```
pleek91, line 158: Love the tests ❤️

src/useEventListener/useEventListener.ts
```diff
@@ -0,0 +1,52 @@
...
+  return {
+    add: () => {
+      addEventListener()
+    },
+    remove: () => {
+      manualMode.value = true
+      removeEventListener()
+    },
+  }
+}
```
pleek91, line 52: Nice, yeah this is really good

pleek91 (approved):
Couple questions about the tests but overall looks really good to me 👍

src/useEventListener/useEventListener.spec.ts
```diff
@@ -0,0 +1,158 @@
...
+    const callback = vi.fn()
+    const target = ref<HTMLElement | undefined | null>(initialValue)
+    vi.spyOn(utils, 'toValue').mockReturnValue(initialValue)
+
+    useEventListener(target, 'click', callback)
+
+    if (target.value) {
+      const addEventListenerMock = vi.spyOn(target.value, 'addEventListener')
+      expect(addEventListenerMock).not.toBeCalled()
+    }
```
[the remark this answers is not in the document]
stackoverfloweth: yeah, this was a bit awkward because the test is currently setup in a way where this would never be executed but it felt like a way of explaining functionality. I don't think it provides much value, I will remove

src/useEventListener/useEventListener.spec.ts
```diff
@@ -0,0 +1,158 @@
...
+    vi.restoreAllMocks()
+  })
+
+  test.each<null | undefined>([
+    undefined,
+    null,
+  ])('given falsy target never adds event listener', (initialValue) => {
+    const callback = vi.fn()
+    const target = ref<HTMLElement | undefined | null>(initialValue)
+    vi.spyOn(utils, 'toValue').mockReturnValue(initialValue)
```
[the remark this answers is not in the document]
stackoverfloweth: wow, this is so obvious in retrospect. Check out the updated tests, I think you'll find them significantly more concise

pleek91 (approved):
❤️

PR #294, opened 2023-08-18
add usePositionStickyObserver

brandonreid:
I still need to test this, but putting it up for early review.


src/useIsStickyStuck/useIsStickyStuck.ts
```diff
@@ -0,0 +1,36 @@
...
+import { MaybeRef } from '@/types/maybe'
+import { useIntersectionObserver } from '@/useIntersectionObserver'
+
+type UseIsStickyStuckOptions = {
+  rootMargin?: string,
+  boundingElement?: HTMLElement | Ref<HTMLElement | undefined>,
+}
+
+export function useIsStickyStuck(
+  element: HTMLElement | Ref<HTMLElement | undefined>,
```
pleek91, line 11: Trying to have all new compositions support getters. 
```suggestion
  element: MaybeRefOrGetter<HTMLElement | undefined>,
```

src/useIsStickyStuck/useIsStickyStuck.ts
```diff
@@ -0,0 +1,36 @@
...
+  options?: MaybeRef<UseIsStickyStuckOptions>,
+): Ref<boolean> {
```
pleek91, line 13: Generally we try and have compositions return objects so in the future we can enhance things without making breaking changes. Can you update this so the syntax is `const { suck } = useIsStickyStuck()`?

src/useIsStickyStuck/useIsStickyStuck.ts
```diff
@@ -0,0 +1,36 @@
...
+  const elementRef = ref(element)
+  const optionsRef = ref(options)
+  const rootMarginRef = ref(optionsRef.value?.rootMargin ?? '-1px 0px 0px 0px')
+  const boundingElementRef = ref(optionsRef.value?.boundingElement ?? document.body)
```
pleek91, line 17: These will need to be computed and react to changes to these values.

src/useIsStickyStuck/useIsStickyStuck.ts
```diff
@@ -0,0 +1,36 @@
...
+import { useIntersectionObserver } from '@/useIntersectionObserver'
+
+type UseIsStickyStuckOptions = {
+  rootMargin?: string,
+  boundingElement?: HTMLElement | Ref<HTMLElement | undefined>,
+}
+
+export function useIsStickyStuck(
+  element: HTMLElement | Ref<HTMLElement | undefined>,
+  options?: MaybeRef<UseIsStickyStuckOptions>,
```
pleek91, line 12: Same with this one

src/useIsStickyStuck/useIsStickyStuck.ts
```diff
@@ -0,0 +1,36 @@
+import { Ref, computed, onMounted, ref } from 'vue'
+import { MaybeRef } from '@/types/maybe'
+import { useIntersectionObserver } from '@/useIntersectionObserver'
+
+type UseIsStickyStuckOptions = {
+  rootMargin?: string,
+  boundingElement?: HTMLElement | Ref<HTMLElement | undefined>,
+}
+
+export function useIsStickyStuck(
```
pleek91, line 10: I don't love this name but I couldn't think of anything better. So I asked chat-GPT! My favorite is `useStuckness` 😂. I don't if you did the same since `useIsStickyStuck` appears in my list as well. I personally like  `usePositionStickyObserver` (not in this list but inspired by the list) which fits nicely with our other "useSomethingObserver" compositions. 

useStickyState
useIsElementStuck
useStickyPositionCheck
useStuckStatus
useStickyElementWatcher
useStickyRefState
useIsStickyStuck
useCheckStickyStatus
useStickyPositionObserver
useStickyElementState
useStickySituation
useStuckElementDetector
useStickyStatusMonitor
useStickyPositionTracker
useStuckness
useIsRefStuck
useStickyRefWatcher
useStuckStateObserver
useStickyElementChecker
useStuckPositionMonitor

brandonreid: Ah `usePositionStickyObserver` for the win for sure, especially when returning an object that we might provide more observation values on in the future.

src/usePositionStickyObserver/usePositionStickyObserver.ts
```diff
@@ -0,0 +1,58 @@
...
+  const stuck = ref(false)
+
+  function intersect(entries: IntersectionObserverEntry[]): void {
+    entries.forEach(entry => {
+      stuck.value = entry.intersectionRatio < 1
+    })
+  }
+
+  const { observe } = useIntersectionObserver(intersect, computed(() => ({
+    threshold: [1],
```
znicholasbrown, line 48: Does a hardcoded threshold make sense here? Asking cause I'm not sure how exactly it's used

src/usePositionStickyObserver/usePositionStickyObserver.ts
```diff
@@ -0,0 +1,58 @@
...
+import { useIntersectionObserver } from '@/useIntersectionObserver'
+import { toValue } from '@/utilities/vue'
+
+export type UsePositionStickyObserverResponse = {
+  stuck: Ref<boolean>,
+}
+
+export type UsePositionStickyObserverOptions = {
+  rootMargin?: string,
+  boundingElement?: MaybeRefOrGetter<HTMLElement | undefined>,
```
pleek91, line 12: I'd remove the MaybeRrfOrGetter type from this. If anyone needs to use a ref or getter they can pass a getter for the options themselves. 

Might simplify the options computed below. 

But doesn't hurt to keep it if you think it's needed.

src/usePositionStickyObserver/usePositionStickyObserver.ts
```diff
@@ -0,0 +1,58 @@
...
+  rootMargin: '-1px 0px 0px 0px',
+  boundingElement: document.body,
+}
+
+export function usePositionStickyObserver(
+  element: MaybeRefOrGetter<HTMLElement | undefined>,
+  options?: MaybeRefOrGetter<UsePositionStickyObserverOptions>,
+): UsePositionStickyObserverResponse {
+  const elementRef = computed(() => toValue(element))
+  const optionsRef = computed(() => {
```
pleek91, line 25: Rather than creating this computed and then a second computed for the visibility observer, I'd just have this return the type the visibility observer expects. That way we don't need two computeds

pleek91 (approved):
Couple thoughts. But overall what I'd expect. Approving so you can move on if this is blocking other PRs.

src/usePositionStickyObserver/usePositionStickyObserver.ts
brandonreid: Really great calls. Updated! Will go ahead and merge to plow ahead, if you look back and catch anything else np.

PR #306, opened 2023-09-14
Enhancement: Run subscription actions in an effect scope that is disposed when a channel is deleted

pleek91:
# Description
Two minor changes
- Use `getCurrentScope` and `onScopeDispose` rather than `getCurrentInstance` and `onUnmounted` in the `useSubscription` composition. This allows subscriptions to automatically unsubscribe when run in any scope and not just components
- Adds an effect scope to `Channel` that is used when running the action. The scope is disposed of when the channel is deleted. 

## Motivation
Automatically wrapping actions in an effect scope means actions can create reactive effects that are automatically disposed. Nested subscriptions will automatically unsubscribe when the parent action is unsubscribed. And any other composition or utility can automatically disposes.

### DIFF
--- a/src/useSubscription/models/channel.ts
+++ b/src/useSubscription/models/channel.ts
@@ -1,5 +1,6 @@
 /* eslint-disable max-classes-per-file */
 import { isEqual } from 'lodash'
+import { effectScope } from 'vue'
 import Manager from '@/useSubscription/models/manager'
 import Subscription from '@/useSubscription/models/subscription'
 import {
@@ -43,6 +44,7 @@ export default class Channel<T extends Action = Action> {
   private readonly action: T
   private readonly args: ActionArguments<T>
   private readonly subscriptions: Map<number, Subscription<T>> = new Map()
+  private readonly scope = effectScope()
   private timer: ReturnType<typeof setInterval> | null = null
   private lastExecution: number = 0
   private _loading: boolean = false
@@ -163,6 +165,7 @@ export default class Channel<T extends Action = Action> {
 
     if (this.subscriptions.size == 0) {
       this.manager.deleteChannel(this.signature)
+      this.scope.stop()
     }
   }
 
@@ -175,7 +178,7 @@ export default class Channel<T extends Action = Action> {
     this.setInterval()
 
     try {
-      this.response = await this.action(...args)
+      this.response = await this.scope.run(() => this.action(...args))
       this.errored = false
       this.error = null
     } catch (error) {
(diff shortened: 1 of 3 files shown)


znicholasbrown (approved):

src/useSubscription/models/channel.ts
```diff
@@ -43,6 +44,7 @@ export default class Channel<T extends Action = Action> {
   private readonly action: T
   private readonly args: ActionArguments<T>
   private readonly subscriptions: Map<number, Subscription<T>> = new Map()
+  private readonly scope = effectScope()
```
stackoverfloweth, line 47: I guess I expected scope to exist per subscription, not per channel. I don't see how this would cascade through children subscriptions if parent scope gets disposed?

src/useSubscription/models/channel.ts
```diff
@@ -163,6 +165,7 @@ export default class Channel<T extends Action = Action> {
 
     if (this.subscriptions.size == 0) {
       this.manager.deleteChannel(this.signature)
+      this.scope.stop()
```
stackoverfloweth, line 168: if children subscriptions aren't properly disposed with parent, it could break this logic to only cleanup channel with all subscriptions are closed

stackoverfloweth (commented):
I might have to read up on how the scope works, this seems too simple to solve such a potentially complex problem 🤔

src/useSubscription/models/channel.ts
pleek91: I think your misunderstanding is how subscriptions and channels are related. The channel executes the action and also cleans up after the action. Each subscription just gets the value of the action and helps inform the channel when the actions should be run.

stackoverfloweth: ah yes, I think I mixed up channel with manager

src/useSubscription/models/channel.ts
pleek91: I'm not following this comment. The action is executed by the channel. And the channel remains "open" until the last subscription unsubscribes. So when the channel is being deleted the scope is stopped which will trigger the dispose in any effects (the child subscriptions).

src/useSubscription/useSubscription.spec.ts
```diff
@@ -555,4 +555,40 @@ describe('subscribe', () => {
...
+      return number
+    }, [number], { manager })
+
+    const spy = vi.spyOn(childSubscription!, 'unsubscribe')
+
+    number.value++
+
+    await timeout()
+
+    expect(spy).toBeCalledTimes(1)
```
stackoverfloweth, line 591: I think it's obvious, but just to confirm we expect child to be unsubscribed _and replaced with a new subscription_?

pleek91: I don't think it will be replaced? Its just unsubscribed. The only way it would get replaced is if the args to that subscription changed.

PR #307, opened 2023-09-18
Enhancement: Create a new scope for each execution when refreshing subscriptions

pleek91:
# Description
When refreshing a subscription (either manually or via an interval) the effect scope of the channel should be disposed. This makes sure any reactive effects are disposed of properly and ensures a refresh executes nested subscriptions.

### DIFF
--- a/src/useSubscription/models/channel.ts
+++ b/src/useSubscription/models/channel.ts
@@ -194,11 +194,9 @@ export default class Channel<T extends Action = Action> {
   }
 
   public refresh(): Promise<void> {
-    const previousScope = this.scope
+    this.scope.stop()
     this.scope = effectScope()
     const response = this.execute()
-
-    previousScope.stop()
 
     return response
   }
--- a/src/useSubscription/models/channel.ts
+++ b/src/useSubscription/models/channel.ts
@@ -44,7 +44,7 @@ export default class Channel<T extends Action = Action> {
   private readonly action: T
   private readonly args: ActionArguments<T>
   private readonly subscriptions: Map<number, Subscription<T>> = new Map()
-  private readonly scope = effectScope()
+  private scope = effectScope()
   private timer: ReturnType<typeof setInterval> | null = null
   private lastExecution: number = 0
   private _loading: boolean = false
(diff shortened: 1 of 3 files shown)


src/useSubscription/models/channel.ts
```diff
@@ -193,6 +193,14 @@ export default class Channel<T extends Action = Action> {
     }
   }
 
+  public refresh(): Promise<void> {
+    this.scope.stop()
+    this.scope = effectScope()
```
stackoverfloweth, line 198: It's unfortunate vue doesn't have a way to "restart" scope, keeping this property readonly

src/useSubscription/useSubscription.spec.ts
```diff
@@ -591,4 +591,20 @@ describe('subscribe', () => {
...
+      useSubscription(action, [], { manager })
+    }, [], { manager })
+
+    await timeout()
+
+    subscription.refresh()
+
+    await timeout()
+
+    expect(action).toBeCalledTimes(2)
```
stackoverfloweth, line 607: 🎉

src/useSubscription/models/channel.ts
```diff
@@ -193,6 +193,14 @@ export default class Channel<T extends Action = Action> {
     }
   }
 
+  public refresh(): Promise<void> {
+    this.scope.stop()
```
stackoverfloweth, line 197: Not even sure it returns a promise, but should this be awaited?

stackoverfloweth (approved):
Super simple solution to a complicated problem.

src/useSubscription/models/channel.ts
pleek91: I also wondered that but no its a sync operation
https://github.com/vuejs/core/blob/main/packages/reactivity/src/effectScope.ts#L81

src/useSubscription/models/channel.ts
pleek91: Yeah I also wanted it to work that way.

znicholasbrown (approved):

PR #311, opened 2023-09-20
Enhancement: Add `onError` callback to subscription options

pleek91:
# Description
Allows an error callback to be provided at implementation. This is a much better method for error processing than checking/watching the `error` and `errored` properties.

### DIFF
--- a/src/useSubscription/models/channel.ts
+++ b/src/useSubscription/models/channel.ts
@@ -68,8 +68,6 @@ export default class Channel<T extends Action = Action> {
     return Math.min(...intervals)
   }
 
-  // conflicting rules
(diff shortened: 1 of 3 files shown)


src/useSubscription/models/channel.ts
```diff
@@ -68,8 +68,6 @@ export default class Channel<T extends Action = Action> {
     return Math.min(...intervals)
   }
 
-  // conflicting rules
-  // eslint-disable-next-line @typescript-eslint/member-ordering
```
pleek91, line 72: Removed these because they're no longer needed

brandonreid (approved):

PR #322, opened 2023-10-05
ensure default value is used if parsed value is nullish

stackoverfloweth:
when the key doesn't exist, the value in storage.ts:32

```ts
const value = storage.getItem(key)
```

comes back as `"null"`, which doesn't get caught by `if (value === null) {` on line 34

this PR ensures that if the value is nullish AFTER parsing, the defaultValue is used

### DIFF
--- a/src/useStorage/storage.ts
+++ b/src/useStorage/storage.ts
@@ -36,7 +36,7 @@ export class StorageManager {
     }
 
     try {
-      return JSON.parse(value) as T
+      return JSON.parse(value) as T | null ?? defaultValue
     } catch {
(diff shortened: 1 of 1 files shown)


pleek91:
If storage.getItem is returning the string `"null"` that means that the value exists in storage doesn't it? It would return an actual `null` if the key does not exist.

stackoverfloweth:
> If storage.getItem is returning the string `"null"` that means that the value exists in storage doesn't it? It would return an actual `null` if the key does not exist.

I could have sworn I did not have a value at the time of testing but when I explicitly delete the key with latest vue-compositions I can not reproduce 🤷 

feel free to close

PR #337, opened 2023-11-01
`useValidation` and `useValidationObserver` - Reset validations

collincchoy:
## The Problem
`useValidation` and `useValidationObserver` are lazy by default and don't trigger until their watched value changes. This is convenient so that initially, the value(s) can be "invalid" but validations have yet to run so there are no errors like in the case of an empty form with required fields. 

However in the same example of having a form with required fields, once the values have been modified, there isn't currently a way to reset the internal `useValidation` (and consequently `useValidationObserver`) state to reset the form back to its initial state (without any validation errors). 

## The Proposed Solution
This PR adds a `reset` method to both useValidation and useValidationObserver. Both methods have the same signature where the latter simply loops over all registered validations and applies the respective reset with parameters drilled down.

It looks like:
```typescript
const value = ref(0)
const { valid, error, reset, validate } = useValidation(value, isEvenNumber)
console.log(valid)  // true

value.value = 1
await validate()
console.log(valid)  // false

reset()
console.log(valid)  // true
```

## An extra wrinkle
So resetting the internal state is great, but what about when the value is reset also? Well because `useValidation` watches the passed in value and automatically re-validates on change, just calling `reset()` like above isn't enough. 

Consider the following code:
```typescript
const value = ref<number | undefined>(undefined)
const { valid, error, reset, validate } = useValidation(value, isEvenNumber)
console.log(valid)  // true
await validate()
console.log(valid)  // false because value is undefined

value.value = 1
await validate()
console.log(valid)  // false

reset()
console.log(valid)  // true but `value.value` is still 1
value.value = undefined
console.log(valid)  // false because validations reran automatically on change
console.log(error)  // non-empty error message
```
When used for form validations, on resetting the form, the above code ends up with an extra round of validations done on the reset value which could add/leave validation messages onscreen.

To get around this, I've added 2 more exposed functions to `useValidation` - `pause` and `resume`. When called, the automatic validation on value change is temporarily paused/resumed respectively. Additionally, the reset method can take an optional callback parameter which runs in between a `pause` and `resume` block so that a consumer can run code without triggering the change effect. 

```typescript
const value = ref<number | undefined>(undefined)
const { valid, error, reset, validate } = useValidation(value, isEvenNumber)
console.log(valid)  // true
await validate()
console.log(valid)  // false because value is undefined

value.value = 1
await validate()
console.log(valid)  // false

reset(() => value.value = undefined)
console.log(valid)  // true
console.log(error)  // empty ""
```

> **Warning**
> The rest of this description is OUTDATED. It's been left here for historical purposes. Changes were made in an attempt to smooth out some of the "sharp edges". Namely that the `reset` method can now take a callback function which should automatically run within a paused validation context and resume afterwards. 

<strike>To get around this, I've added an opt-in flag to the reset method which skips the next on-change-validation execution to allow a consumer to update the value once without triggering validations.

```typescript
const value = ref<number | undefined>(undefined)
const { valid, error, reset, validate } = useValidation(value, isEvenNumber)
console.log(valid)  // true
await validate()
console.log(valid)  // false because value is undefined

value.value = 1
await validate()
console.log(valid)  // false

reset({ skipNextValidateOnChange: true })
console.log(valid)  // true but `value.value` is still 1
value.value = undefined
console.log(valid)  // true
console.log(error)  // empty ''
```

## The sharp edges
On the one hand this now gives a consumer the ability to reset validations and supports the use-case I outlined above for resetting the value along with any related validation state.

On the other hand, this seems a little clunky of an api and also there's potential for race conditions if done out-of-order -- `reset` should ideally be called _before_ the values themselves are updated otherwise the `skipNextValidateOnChange` might not be in effect by the time the watcher's effect triggers.
</strike>

### DIFF
--- a/src/useValidation/useValidation.spec.ts
+++ b/src/useValidation/useValidation.spec.ts
@@ -45,12 +45,12 @@ describe('useValidation', () => {
 
     it('triggers automatically when the value changes', async () => {
       const theValue = ref(0)
-      const validationSpy = vi.fn().mockResolvedValue('Validation did not pass')
+      const validationRule = vi.fn().mockResolvedValue('Validation did not pass')
       const wrapper = mount({
         setup() {
           const { valid, error } = useValidation(
             theValue,
-            validationSpy,
+            validationRule,
           )
 
           return { valid, error }
@@ -59,13 +59,13 @@ describe('useValidation', () => {
 
       expect(wrapper.vm.valid).toBe(true)
       expect(wrapper.vm.error).toBe('')
-      expect(validationSpy).not.toHaveBeenCalled()
+      expect(validationRule).not.toHaveBeenCalled()
 
       theValue.value = 1
       await nextTick()
       await flushPromises()
 
-      expect(validationSpy).toHaveBeenCalled()
+      expect(validationRule).toHaveBeenCalled()
       expect(wrapper.vm.error).toBe('Validation did not pass')
       expect(wrapper.vm.valid).toBe(false)
     })
--- a/src/useValidation/useValidation.spec.ts
+++ b/src/useValidation/useValidation.spec.ts
@@ -17,7 +17,6 @@ describe('useValidation', () => {
     it('sets valid to true when the rules pass', async () => {
       const { valid, error, validate } = useValidation(
         ref(1),
-        'Number',
         isGreaterThanZero,
       )
 
@@ -28,16 +27,17 @@ describe('useValidation', () => {
     })
 
     it('sets valid to false with an error message when the rules do not pass', async () => {
+      const errorMessage = 'Validation did not pass'
+      const validationRule: ValidationRule<number> = () => errorMessage
       const { valid, error, validate } = useValidation(
         ref(0),
-        'Number',
-        isGreaterThanZero,
+        validationRule,
       )
 
       await validate()
 
       expect(valid.value).toBe(false)
-      expect(error.value).toBe('Number must be greater than 0')
+      expect(error.value).toBe(errorMessage)
     })
 
     it('triggers automatically when the value changes', async () => {
@@ -111,14 +111,13 @@ describe('useValidation', () => {
     it('resets the validation state', async () => {
       const { state, validate, reset } = useValidation(
         ref(0),
-        'Number',
-        isGreaterThanZero,
+        () => 'Validation did not pass',
       )
 
       await validate()
 
       expect(state.valid).toBe(false)
-      expect(state.error).toBe('Number must be greater than 0')
+      expect(state.error).toBe('Validation did not pass')
 
       reset()
 
--- a/src/useValidation/useValidation.spec.ts
+++ b/src/useValidation/useValidation.spec.ts
@@ -170,5 +170,42 @@ describe('useValidation', () => {
       await updateTheValueAndWaitForEffects(-1)
       expect(wrapper.vm.state.valid).toBe(false)
     })
+
+    it('resumes validation if an error occurs in the reset callback', async () => {
+      const theValue = ref(0)
+      const validationRule = vi.fn().mockResolvedValue('Validation did not pass')
+      const wrapper = mount({
+        setup() {
+          const { state, reset } = useValidation(
+            theValue,
+            validationRule,
+          )
+
+          return { state, reset }
+        },
+      })
+
+      expect(wrapper.vm.state.valid).toBe(true)
+      expect(wrapper.vm.state.error).toBe('')
+      expect(validationRule).not.toHaveBeenCalled()
+
+      try {
+        wrapper.vm.reset(() => {
+          theValue.value = 1
+          throw new Error('Something went wrong')
+        })
+      } catch (error) {
+        // ignore
+      }
+      await flushPromises()
+
+      expect(validationRule).not.toHaveBeenCalled()
+      expect(wrapper.vm.state.valid).toBe(true)
+
+      // watcher resumed
+      theValue.value = 2
+      await flushPromises()
+      expect(validationRule).toHaveBeenCalled()
+    })
   })
 })
--- a/src/useValidation/useValidation.spec.ts
+++ b/src/useValidation/useValidation.spec.ts
@@ -14,9 +14,12 @@ const isGreaterThanZero: ValidationRule<number> = (value, name) => {
 
 describe('useValidation', () => {
   describe('validate', () => {
-    it('sets valid to true when the rules pass', async () => {
+    it.each([
+      { value: 1, valueType: 'plain value' },
+      { value: ref(1), valueType: 'ref' },
+    ])('sets valid to true when the rules pass (with $valueType)', async ({ value }) => {
       const { valid, error, validate } = useValidation(
-        ref(1),
+        value,
         isGreaterThanZero,
       )
 
@@ -26,11 +29,14 @@ describe('useValidation', () => {
       expect(error.value).toBe('')
     })
 
-    it('sets valid to false with an error message when the rules do not pass', async () => {
+    it.each([
+      { value: 0, valueType: 'plain value' },
+      { value: ref(0), valueType: 'ref' },
+    ])('sets valid to false with an error message when the rules do not pass (with $valueType)', async ({ value }) => {
       const errorMessage = 'Validation did not pass'
       const validationRule: ValidationRule<number> = () => errorMessage
       const { valid, error, validate } = useValidation(
-        ref(0),
+        value,
         validationRule,
       )
 
--- a/src/useValidation/useValidation.spec.ts
+++ b/src/useValidation/useValidation.spec.ts
@@ -1,0 +1,41 @@
+import { describe, expect, it } from 'vitest'
+import { ref } from 'vue'
+import { ValidationRule, useValidation } from '@/useValidation/useValidation'
+
+const isGreaterThanZero: ValidationRule<number> = (value, name) => {
+  if (value > 0) {
+    return true
+  }
+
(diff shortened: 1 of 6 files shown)


pleek91:
@collincchoy haven't looked at the code yet. But one question. What happens if you call reset on the validation AFTER you reset the form? That is the order I expected. And seems like that would eliminate the need for a validation skip argument. But I'm guessing that has some other side effect

collincchoy:
> @collincchoy haven't looked at the code yet. But one question. What happens if you call reset on the validation AFTER you reset the form? That is the order I expected. And seems like that would eliminate the need for a validation skip argument. But I'm guessing that has some other side effect

Timing issue, the reset could happen before the watch effect hits. Although maybe that can be handled as part of the resetting 🤔

pleek91:
> > @collincchoy haven't looked at the code yet. But one question. What happens if you call reset on the validation AFTER you reset the form? That is the order I expected. And seems like that would eliminate the need for a validation skip argument. But I'm guessing that has some other side effect
> 
> Timing issue, the reset could happen before the watch effect hits("could" as in that seems to be what I'm observing in practice). Although maybe that can be handled as part of the resetting ~🤔~
> 
> edit: I thought I could do something w/ the pending state but my idea didn't work out.

@collincchoy interesting. What do you think about adding some sort of stop/start functionality along with reset? I'm not to keen to introduce `skipNextValidateOnChange`. But this devx seems like it would be useful. Its more verbose but gives control to implementation. Its also a useful feature on its own.
```typescript
const value = ref(0)
const { reset, pause, resume } = useValidation(value, isEvenNumber)

reset()
pause()
value.value = 0
resume()
```

collincchoy:
> > > @collincchoy haven't looked at the code yet. But one question. What happens if you call reset on the validation AFTER you reset the form? That is the order I expected. And seems like that would eliminate the need for a validation skip argument. But I'm guessing that has some other side effect
> > 
> > 
> > Timing issue, the reset could happen before the watch effect hits("could" as in that seems to be what I'm observing in practice). Although maybe that can be handled as part of the resetting ~🤔~
> > edit: I thought I could do something w/ the pending state but my idea didn't work out.
> 
> @collincchoy interesting. What do you think about adding some sort of stop/start functionality along with reset? I'm not to keen to introduce `skipNextValidateOnChange`. But this devx seems like it would be useful. Its more verbose but gives control to implementation. Its also a useful feature on its own.
> 
> ```ts
> const value = ref(0)
> const { reset, pause, resume } = useValidation(value, isEvenNumber)
> 
> reset()
> pause()
> value.value = 0
> resume()
> ```

Cool I think those would be useful levers to pull. what if reset took an optional callback that runs within a pause/resume block?

pleek91:
We talked about this already but commenting here for future us. I really like the idea of a callback in the reset function that allows you to do some stuff while resetting 👍

src/useValidationObserver/useValidationObserver.spec.ts
```diff
@@ -0,0 +1,133 @@
...
+    const result = await wrapper.vm.validate()
+
+    expect(result).toBe(true)
+    expect(validationRule1).toHaveBeenCalled()
+    expect(validationRule2).toHaveBeenCalled()
+  })
+
+  it('should observe multiple useValidations and be invalid if any does not pass', async () => {
+    const validationRule1Spy = vi.fn().mockResolvedValue(true)
+    const validationRule2Spy = vi.fn().mockResolvedValue('Number must be greater than 0')
```
pleek91, line 46: You can also just return `false` from a validation rule and it'll give a generic error. So you probably don't need to mock error strings in these unless you're testing the strings specifically.

src/useValidation/useValidation.spec.ts
```diff
@@ -0,0 +1,134 @@
...
+      await validate()
+
+      expect(valid.value).toBe(false)
+      expect(error.value).toBe('Number must be greater than 0')
+    })
+
+    it('triggers automatically when the value changes', async () => {
+      const theValue = ref(0)
+      const validationRule = vi.fn().mockResolvedValue('Validation did not pass')
+      const wrapper = mount({
```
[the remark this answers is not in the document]
collincchoy: The watch effect only kicks in after mount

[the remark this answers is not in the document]
pleek91: ohhh right. Forgot that existed. That's annoying. Carry on

src/useValidation/useValidation.spec.ts
```diff
@@ -0,0 +1,134 @@
...
+        },
+      })
+
+      expect(wrapper.vm.valid).toBe(true)
+      expect(wrapper.vm.error).toBe('')
+      expect(validationRule).not.toHaveBeenCalled()
+
+      theValue.value = 1
+      await nextTick()
+      await flushPromises()
```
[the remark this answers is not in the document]
collincchoy: ah yeah I noticed this after.. a bit. I suppose i should remove all of the nextTick calls since to your point it's not quite right. when it does work, it's probably happen-stance but originally I didn't realize that. In the past i've seen/used [flush-promises](https://www.npmjs.com/package/flush-promises) but rather than add it, I stumbled onto this rather simple implementation (from [here](https://rickschubert.net/blog/posts/flushing-promises/)). 

i didn't see the other timeout utility. Hmmm tbh idk off-hand the difference but I can look into it a bit more and/or remove in favor of the existing timeout utility.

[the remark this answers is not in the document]
pleek91: I was just curious if you knew something about nextTick that I didn't. Cause I can never get that to work like it seems it should based on the description in the vue docs.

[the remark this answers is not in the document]
collincchoy: i really thought nextTick was working for... longer than I expected 😅 (well the tests passed!), but I guess it makes sense that effects run off a separate schedule entirely from the batched DOM updates that nextTick is i guess more for...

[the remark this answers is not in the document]
pleek91: Even when I'm working with the dom nextTick never works lol.

src/useValidation/useValidation.ts
```diff
@@ -22,8 +22,21 @@ export type ValidateMethodOptions = {
 
 export type ValidateMethod = (options?: ValidateMethodOptions) => Promise<boolean>
 
+export type ResetMethodParams = [
+  /**
+   * If true, the next call to validate will not be called.
+   * This allows you to reset validation state and then reset the value
+   * without triggering another validation.
+   */
+  resetCallback?: () => void
```
pleek91, line 31: I think this comment is outdated?

src/useValidationObserver/useValidationObserver.spec.ts
```diff
@@ -0,0 +1,131 @@
...
+  })
+
+  it('should observe multiple useValidations', async () => {
+    const validationRule1 = vi.fn().mockResolvedValue(true)
+    const validationRule2 = vi.fn().mockResolvedValue(true)
+    const wrapper = mount({
+      setup() {
+        const { validate } = useValidationObserver()
+        useValidation(ref(0), validationRule1)
+        useValidation(ref(0), validationRule2)
```
pleek91, line 32: it doesn't matter but for these tests where you're not changing the value you can just pass a static value. It doesn't have to be a ref.

pleek91 (commented):
One question but otherwise this looks great. Especially thankful for all the test coverage. Thank you!

collincchoy:
Some dependency bump broke the `wrapper.vm.*` types.. need to dig in later.

collincchoy:
> Some dependency bump broke the `wrapper.vm.*` types.. need to dig in later.

A-ha. I finally found it. bumping @testing-library/vue seems to have bumped @vue/test-utils to > 2.4.0 where 2.4.0 contained breaking typing changes where previously it was probably at @vue/test-utils@2.3.2-ish.

src/useValidation/useValidation.spec.ts
pleek91, line 1: ❤️

src/useValidation/useValidation.ts
```diff
@@ -98,6 +112,33 @@ export function useValidation<T>(
...
+  const reset: ResetMethod = (resetCallback) => {
+    error.value = ''
+    pending.value = false
+    validated.value = false
+
+    if (resetCallback) {
+      pause()
+      try {
+        resetCallback()
+      } finally {
```
pleek91, line 124: We probably want to catch the error and `console.error` it. Otherwise we'll swallow errors and could make debugging difficult when using this feature.

pleek91 (approved):
This looks great. One error we should log out but otherwise this is ready!

src/useValidation/useValidation.ts
collincchoy: Hm really? 

I think that'd be the case if there was an empty `catch(e) {}` block here but since there's no catch, an error remains uncaught and should bubble up (-- and the finally block should still run as expected)? 🤔 

I think I'd expect to be able to catch an error from the callback with a try/catch wrapped around `reset`.. but hmm I can verify that behavior real quick tho.

pleek91: oh, no I believe you're correct. I wasn't thinking about this correctly.

collincchoy: cool - tweaked this test to make sure it works as expected - [d037165](https://github.com/PrefectHQ/vue-compositions/pull/337/commits/d037165d0a2dd1b43118327a2a663f8cc1ef17cd)

pleek91: TY, that does make it more clear 👍

PR #367, opened 2024-01-03
Add some jsdoc for better IDE devX

collincchoy:
# Description
This PR introduces JSDoc strings in a couple of places to improve devX when consuming the following functions in other projects downstream. 

### `useSubscription`

 https://github.com/PrefectHQ/vue-compositions/assets/22418768/7e4bd17b-80f4-4743-8cb5-b4de9d99a9c0 

### `useSubscriptionWithDependencies`
image

### `useValidation` 
https://github.com/PrefectHQ/vue-compositions/assets/22418768/9a400f4e-9097-4523-ac1d-34671e3ff74e

### `useValidationObserver`
https://github.com/PrefectHQ/vue-compositions/assets/22418768/71c8c621-2a07-4e41-9537-dae100f40e3d

### DIFF
--- a/src/useSubscription/useSubscription.ts
+++ b/src/useSubscription/useSubscription.ts
@@ -32,19 +32,6 @@ const defaultManager = new Manager()
  *
  * @param action - The function to be executed.
  * @param args - Parameters of the action being executed.
- * @param options - An object containing optional parameters.
- * @param options.interval - A max poll interval in milliseconds.
- * @param options.manager - A channel manager to use for the subscription.
- *
- * @returns An object containing the following properties:
- * - `loading`: Set to `true` while the action is being executed.
- * - `errored`: Set to `true` if there is an error when executing the action.
- * - `error`: Stores any error thrown while executing the action.
- * - `response`: Return value from the `action` after execution.
- * - `refresh`: Executes the `action` again.
- * - `promise`: Returns a promise that resolves when the `response` is returned.
- * - `unsubscribe`: Remove the subscription from the channel.
- * - `subscribe`: Resubscribes to the channel using the same arguments.
  */
 export function useSubscription<T extends Action>(
   ...[action, args, optionsArg = {}]: SubscribeArguments<T>
--- a/src/useSubscription/useSubscription.ts
+++ b/src/useSubscription/useSubscription.ts
@@ -9,7 +9,46 @@ import { uniqueValueWatcher } from '@/utilities/uniqueValueWatcher'
 
 const defaultManager = new Manager()
 
-export function useSubscription<T extends Action>(...[action, args, optionsArg = {}]: SubscribeArguments<T>): UseSubscription<T> {
+/**
+ * The `useSubscription` composition manages data sharing across components. Multiple components can subscribe to an `action` (any method or function) and share the response value.
+ *
+ * @example
+ * ```typescript
+ * import { useSubscription } from '@prefecthq/vue-compositions'
+ *
+ * const subscription = useSubscription(action, args, options)
+ * ```
+ *
+ * ## How it works
+ * When creating a subscription for the first time a `Channel` is created. A channel is unique to both the `action` and the `args` used to create the subscription. When a channel is created it immediately executes the action with the given args. If `options.interval` is set the action will execute again at that interval updating the `loading`, `errored`, `error`, and `response` values automatically.
+ *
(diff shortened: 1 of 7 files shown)


src/useSubscription/useSubscription.ts
```diff
@@ -9,7 +9,33 @@ import { uniqueValueWatcher } from '@/utilities/uniqueValueWatcher'
...
+ *
+ * If another subscription is created with the same action and args, it will not execute the action again. Instead the same channel will be reused. So if multiple components subscribe using the same args the action will only be executed once.
+ *
+ * If any subscriptions have `options.interval` then the shortest interval will be used. All subscriptions will received an updated response at the decided interval, even if a specific subscription has no interval or if it has a greater interval. Interval is basically "maximum age of execution" which defaults to `Infinity`.
+ *
+ * When a subscription is removed the channel interval is recalculated. A subscription can be removed by calling `subscription.unsubscribe()` or the subscription will automatically be removed when the component that created the subscription is unmounted.
```
pleek91, line 29: This is really long and also duplicated from the readme. Can we make a short summary? That way we don't have to make sure this and the readme is in sync

src/useSubscription/useSubscription.ts
```diff
@@ -9,7 +9,33 @@ import { uniqueValueWatcher } from '@/utilities/uniqueValueWatcher'
 
 const defaultManager = new Manager()
 
-export function useSubscription<T extends Action>(...[action, args, optionsArg = {}]: SubscribeArguments<T>): UseSubscription<T> {
+/**
+ * The `useSubscription` composition manages data sharing across components. Multiple components can subscribe to an `action` (any method or function) and share the response value.
```
collincchoy, line 13: From the README minus the top heading and the params/return value tables at the end. 

Considered dropping the README in favor of these more closely co-located documentation bits for easier long-term maintenance, but left it in for now and am using it as a `@see` link both here and for `useSubscriptionWithDependencies` since symbol linking only works if the symbol is imported into the current module so linking to the static readme file is a tad easier.

pleek91 (approved):

src/useValidation/useValidation.ts
```diff
@@ -22,25 +22,31 @@ export type ValidateMethodOptions = {
...
+  *   another validation on change.
+  *
+  * @example
+  * ```ts
+  * const value = ref(0)
+  * const { reset } = useValidation(value, isGreaterThanZero)
+  * reset(() => value.value = 0)
+  * ```
+  */
+  // eslint-disable-next-line @typescript-eslint/method-signature-style -- JSDoc in IDE doesn't work nearly as well with a property
```
collincchoy, line 41: I played with this a _bunch_ and this was the best I could get to. Adding docs to a separate `ResetMethod` type gets a little lost and there's some finickyness with maintaining the `@param` linking as things get extracted out into TS types - seems like vs code stops tracking it as a function at various points...

src/useSubscription/useSubscription.ts
collincchoy: https://github.com/PrefectHQ/vue-compositions/pull/367/files#r1441095905

Yeah it is duplicated from the README. .. 🤔 

I'm not sure that I think the length is an _issue_ tho. I think it's near the maximum length we probably want but personally I really like how much more accessible this would make this information. Agreed keeping this and the readme in-sync sounds 👎 . 

IMO this is preferable to the README but i wasn't sure about outright deleting it myself. 

All that said tho, I'm more than happy to defer to you on this. Even for a more concise description here - do you have any particular suggestions as you definitely understand this thing better than me 😅 . Or if you'd prefer to keep the README as the source of truth, what do you think about just

```js
/**
 * The `useSubscription` composition manages data sharing across components. Multiple components can subscribe to an `action` (any method or function) and share the response value.
 *
 * @see [README](https://github.com/PrefectHQ/vue-compositions/tree/main/src/useSubscription#readme)
 *
 * @param action - The function to be executed.
 * @param args - Parameters of the action being executed.
 */
```

pleek91: Yeah that snippet preferable to me. I think keeping the readmes is necessary since the whole repository is explained by readmes.

collincchoy: Ah I just realized that'd still be duplicating the docs for params/return values since those are inlined over in `susbscription.ts`. Defining them via @param/@return here doesn't work with the extracted return type and also doesn't play as well with destructured return values.

collincchoy: Updated in b4a8a53. 👍

PR #374, opened 2024-01-17
Add Devtools for `useSubscription`

collincchoy:
This PR adds a Vue Devtools plugin for debugging `useSubscription`. 

Enable in-app with:
```typescript
import { plugin as VueCompositionsDevtools } from '@prefecthq/vue-compositions'
...

const app = createApp(App)
if (APP_MODE() === 'development' || DEPLOYMENT_ENVIRONMENT() !== 'production') {
  app.use(VueCompositionsDevtools)
}
```

### Custom inspector
image

### Custom timeline layer and events
image


src/useSubscription/useSubscriptionDevtools.ts
```diff
@@ -0,0 +1,243 @@
...
+  //   label: 'Timeline Color',
+  //   type: 'choice',
+  //   defaultValue: 0xFF69B4,
+  //   options: [
+  //     { value: 0x4fc08d, label: 'Vue Green' },
+  //     { value: 0x175cd3, label: 'Prefect Blue' },
+  //     { value: 0xFF69B4, label: 'Hot' }
+  //   ],
+  //   component: 'button-group'
+  // }
```
collincchoy, line 51: We can add plugin settings like this https://devtools.vuejs.org/plugin/plugins-guide.html#plugin-settings

Here I was going to add a toggle to switch the color on the timeline but... changing the layer's color after having created it already isn't supported... 😢 . I thought i'd leave this scaffolding here tho incase some other setting comes to mind to add in. 

1 thought I had was a giant list of all timeline events so you could filter down which ones get emitted/tracked.

src/useSubscription/useSubscriptionDevtools.ts
```diff
@@ -0,0 +1,243 @@
...
+        key: String(id),
+        value: getComponentName(vm),
+      })
+    ),
+  }
+}
+
+function getComponentName(vm: ComponentInternalInstance | null): string {
+  // @ts-ignore __name is not in the types but works.
+  return vm?.type.__name as string ?? vm?.type?.name ?? '🤷🏻‍♂️ Unknown component'
```
collincchoy, line 197: there are some cases where I haven't been able to get the actual component name... I think this might be due to some package linking issues... for example, components in ui-library are built into a prod dist and lose their debug info. or something else entirely i haven't been able to track down. 

There is commented code further up which pulls the component names from the vue devtools plugin api - it also doesn't seem to capture all names. Not sure if it's better/worse than this but I'm imagining it'd be slower than just this.

src/useSubscription/useSubscriptionDevtools.ts
```diff
@@ -0,0 +1,243 @@
...
+}
```
pleek91, line 198: That's dumb, you'd think the devtools api at least would expose this somehow.

pleek91 (approved):
Do you mind linting these files? Otherwise ship it!

src/useSubscription/models/channel.ts
```diff
@@ -60,6 +61,12 @@ export default class Channel<T extends Action = Action> {
     this.args = args
   }
 
+  public get actionName(): string {
+    const prefix = "bound "
+    const sanitizedChannelActionName = this.action.name.startsWith(prefix) ? this.action.name.slice(prefix.length) : this.action.name
+    return sanitizedChannelActionName
+  }
```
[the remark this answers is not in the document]
collincchoy: the actions are callables that (are often) bounded(like `theActionFn.bind(...)` so `this.action.name` => `bound getData`. This strips the "bound" prefix if it's there else return `this.action.name`.

package.json
```diff
@@ -68,6 +68,7 @@
     }
   },
   "dependencies": {
+    "@vue/devtools-api": "^6.1.4",
```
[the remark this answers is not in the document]
collincchoy: I had the same thought and similarly _thiiink_ it needs to be non-dev dep so that it's included in the package dist?

src/useSubscription/models/channel.ts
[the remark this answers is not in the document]
pleek91: oh interesting, TIL. TY

package.json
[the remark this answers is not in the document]
pleek91: Yeah you're probably right 🤷

src/useSubscription/useSubscriptionDevtools.ts
pleek91: interesting, I don't feel strongly but I know vue highly discourages using internal properties because they can change without notice. But pretty low risk for devtools so 🤷

collincchoy: I swapped the implementation over in [f4b2832](https://github.com/PrefectHQ/vue-compositions/pull/374/commits/f4b283224274d4b0369aa674ee392ff43302aaf3).

.vscode/settings.json
```diff
@@ -16,6 +16,16 @@
   "editor.codeActionsOnSave": {
     "source.fixAll.eslint": "explicit"
   },
+  "eslint.options": {
```
collincchoy, line 19: I guess I need these for my vs code to run eslint on save...

but now i'm slightly confused b/c. ... how dis work before. 🤷

pleek91: huh that is weird 🤔 but that's what we have in ui-library 🤷

PR #376, opened 2024-01-18
Chore/verify release job

jimid27:
- Relates: https://github.com/PrefectHQ/platform/issues/5517
- Adds a check to verify a package version has been updated before moving forward w/ a release
  - Uses: https://github.com/PrefectHQ/actions-verify-npm-package-bump

Tested: https://github.com/PrefectHQ/vue-charts/actions/runs/7572241849 & https://github.com/PrefectHQ/vue-charts/actions/runs/7572230978

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -13,7 +13,6 @@ jobs:
   verify-release:
     name: Verify Release
     runs-on: ubuntu-latest
-    environment: 'prod'
     outputs:
       release: ${{ steps.verify-release.outputs.release }}
     steps:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -8,12 +8,29 @@ on:
       - main
       
 jobs:
+  # This job is used to verify that the version in package.json does not match the latest tag.
+  # If it does, we do not want to proceed with the release.
+  verify-release:
+    name: Verify Release
+    runs-on: ubuntu-latest
+    environment: 'prod'
+    outputs:
+      release: ${{ steps.verify-release.outputs.release }}
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
 
+      - name: Verify Release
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -8,12 +8,28 @@ on:
...
+      release: ${{ steps.verify-release.outputs.release }}
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
 
+      - name: Verify Release
+        id: verify-release
+        uses: prefecthq/actions-verify-npm-package-bump@main
+  # This job is used to build and publish the release if the verify-release job passes.
```
pleek91, line 26: Can you explain why this is needed? I've gotten a bit removed from how this all works but the release step has always stopped without attempting to publish duplicate packages. Is this because the update PR step isn't aware that no new package was published? The action for that also has an output that says if a publish happened https://github.com/JS-DevTools/npm-publish.

jimid27: It's actually a bit stupid but essentially we run this "release" workflow when two params are met:
1. The package.json file has change
2. A merge to main has occured
The issue with this is that the package.json can change without a bump in the actual package version in the package.json.  So what this job does is verify that a version change has occured before moving onto the release step.  Basically avoids these errors in the job: https://github.com/PrefectHQ/vue-compositions/actions/runs/7563179529/job/20595077477

pleek91: Would it help to change this to run off of the tag that gets pushed with the package change? That would prevent this from being triggered when just a package is updated or something.

jimid27: We def could move to only releasing when a tag is created.  As it stands right now this release job actually creates the tag and release

jimid27: 🤦 I'm way overcomplicating this.  You are 100% correct.  We can just use the output. Updating all now 🤦

PR #377, opened 2024-01-18
Enhancement: Add pause/resume functionality to subscriptions

pleek91:
# Description
Adds the ability for `Manager` to pause and resume all of the active channels. This will allow us to automatically pause making requests when a tab or window has been backgrounded. 

```typescript
import { defaultSubscriptionManager } from '@prefecthq/vue-compositions'

// logic for when pause should happen omitted
defaultSubscriptionManager.pause()
```

When a channel is paused it will not execute the action. If an action execution is requested (either via refresh or an interval poll) the channel is marked as `late`. When a channel is resumed if it missed an execution while it was paused and is in a late state, it will immediately execute the action via a refresh.

Individual subscriptions have a reactive `paused` and `late` property that reflects the state of the channel.

### DIFF
--- a/src/useSubscription/models/channel.ts
+++ b/src/useSubscription/models/channel.ts
@@ -48,7 +48,7 @@ export default class Channel<T extends Action = Action> {
   private scope = effectScope()
   private timer: ReturnType<typeof setInterval> | null = null
   private lastExecution: number = 0
-  private late: boolean = false
+  private _late: boolean = false
   private _paused: boolean = false
   private _loading: boolean = false
   private _executed: boolean = false
@@ -84,8 +84,24 @@ export default class Channel<T extends Action = Action> {
   public set paused(paused: boolean) {
     this._paused = paused
 
+    for (const subscription of this.subscriptions.values()) {
+      subscription.paused.value = paused
+    }
+
     if (this.late) {
       this.execute()
+    }
+  }
+
+  private get late(): boolean {
+    return this._late
+  }
+
+  private set late(late: boolean) {
+    this._late = late
+
+    for (const subscription of this.subscriptions.values()) {
+      subscription.late.value = late
     }
   }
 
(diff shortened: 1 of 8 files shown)


src/useSubscription/models/channel.ts
```diff
@@ -75,6 +77,46 @@ export default class Channel<T extends Action = Action> {
...
+  public set paused(paused: boolean) {
+    this._paused = paused
+
+    for (const subscription of this.subscriptions.values()) {
+      subscription.paused.value = paused
+    }
+
+    if (this.late) {
+      this.refresh()
+    }
```
znicholasbrown, line 93: I'd expect that `refresh` only gets called if the subscription/channel isn't paused but this looks like it'll call the action no matter what when `paused` is set

znicholasbrown: Otherwise I'm not totally sure what `late` is for except informational purposes, i.e. this data is known to be stale because something tried to update it but couldn't

znicholasbrown (commented):
Love that this already has dev tools updates

src/useSubscription/models/channel.ts
pleek91: Hmm right, though it cannot be `late` if it wasn't paused. So I don't think this could ever be called when setting `paused: true`. But I'll add a check to make that more clear 👍

znicholasbrown (approved):

src/useSubscription/models/channel.ts
znicholasbrown: Yeah that makes sense - thanks!

pleek91: Correct, `late` is set to true when the action is executed (either from an interval or a refresh) but the execution was prevented because the channel is paused. So late just informs us that we need to refresh when resuming.

PR #385, opened 2024-02-01
Chore: expose useEventListener

dylanbhughes:
When I tried to use `useEventListener` from an external project, I was unable to import it. Looks like we need to expose it.


pleek91 (approved):
🤦

PR #393, opened 2024-02-20
Subscriptions: Add the ability to refresh a channel without creating a subscription

pleek91:
# Description
Introduced a `refresh` method on `SubscriptionManager` and a `refreshChannel` utility that wraps it. Calling refresh will refresh a channel directly without creating a new subscription. If not channel exists its a noop.


collincchoy (approved):
🎉 ez-peez

PR #398, opened 2024-02-29
Add `maxRefreshRate` directly to `Channel` and `Subscription` classes

pleek91:
# Description
Builds on https://github.com/PrefectHQ/vue-compositions/pull/397 by adding `maxRefreshRate` logic directly to the channel's refresh method itself rather than using it only in `manager.refresh`. Also exposes the option in `subscription.refresh` so you can use it on existing subscriptions

### DIFF
--- a/src/useSubscription/utilities/refresh.ts
+++ b/src/useSubscription/utilities/refresh.ts
@@ -1,11 +1,13 @@
+import { SubscriptionManager } from '@/useSubscription/models/manager'
 import { Action, ActionArguments } from '@/useSubscription/types/action'
 import { RefreshChannelOptions } from '@/useSubscription/types/channels'
 import { defaultSubscriptionManager } from '@/useSubscription/useSubscription'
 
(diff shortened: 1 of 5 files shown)


src/useSubscription/utilities/refresh.ts
```diff
@@ -1,11 +1,13 @@
...
 import { Action, ActionArguments } from '@/useSubscription/types/action'
 import { RefreshChannelOptions } from '@/useSubscription/types/channels'
 import { defaultSubscriptionManager } from '@/useSubscription/useSubscription'
 
+
 export function refreshChannel<T extends Action>(
   action: T,
   args: ActionArguments<T>,
-  options?: RefreshChannelOptions,
+  options?: RefreshChannelOptions & { manager?: SubscriptionManager },
```
collincchoy, line 10: 🎉

collincchoy (approved):

PR #404, opened 2024-03-15
Feature: `useMousePosition`

znicholasbrown:
Adds a new composition, `useMousePosition` used to passively track the position of the mouse without needing to create/tear down global event listeners.


pleek91 (approved):
Nice!

PR #406, opened 2024-03-19
Chore: Add a test for refresh channel

pleek91:
# Description
I was testing this locally and realized a test case was missing.


zhen0 (approved):
The linter has what looks like a legit an issue but otherwise looks good. Approving as you can decide if the issue needs addressing.

PR #415, opened 2024-04-02
Feature: `useClickOutside` 

znicholasbrown:
Adds a composition to consolidate some logic around click handling for things like modals/popups/tooltips etc.


pleek91 (approved):
Nice

PR #418, opened 2024-04-02
[useValidation] Add support for getters for value in useValidation.

collincchoy:


src/useValidation/useValidation.ts
```diff
@@ -61,10 +61,10 @@ function isRules<T>(value: MaybeRef<string> | RulesArg<T>): value is RulesArg<T>
...
 }
 
-export function useValidation<T>(value: MaybeRef<T>, rules: RulesArg<NoInfer<T>>): UseValidation
-export function useValidation<T>(value: MaybeRef<T>, name: MaybeRef<string>, rules: RulesArg<NoInfer<T>>): UseValidation
+export function useValidation<T>(value: MaybeRefOrGetter<T>, rules: RulesArg<NoInfer<T>>): UseValidation
+export function useValidation<T>(value: MaybeRefOrGetter<T>, name: MaybeRef<string>, rules: RulesArg<NoInfer<T>>): UseValidation
 export function useValidation<T>(
-  value: MaybeRef<T>,
+  value: MaybeRefOrGetter<T>,
   nameOrRules: MaybeRef<string> | RulesArg<NoInfer<T>>,
```
collincchoy, line 68: I looked into trying to make name be passable via a getter as well, but it gets more complicated trying to distinguish between a rule (which can be a function) and a getter.

src/useValidation/useValidation.ts
```diff
@@ -77,7 +77,7 @@ export function useValidation<T>(
     throw new Error('Invalid useValidation arguments')
   }
 
-  const valueRef = ref(value) as Ref<T>
+  const valueRef = toRef(value)
```
pleek91, line 80: ayyy nice got rid of an `as`

src/useValidation/useValidation.ts
pleek91: yeah probably not worth it right now.

pleek91 (approved):

PR #419, opened 2024-04-02
Remove some 3.3. shims

collincchoy:
Now that vue-comp has a peer dependency of vue@^3.3.0, we can remove:
- toValue
- MaybeRef
- MaybeRefOrGetter

as they're all built-in to vue now.


pleek91 (approved):
Yessss ❤️

PR #425, opened 2024-04-09
Bugfix: `useRouteQueryParam` over-eager reactive updates

znicholasbrown:
This PR introduces a new intermediary watcher for `useRouteQueryParam`. This watcher performs the task of watching the query object for the passed key and updating an internal reactive value instead of passing back a computed getter that reacts to the entire reactive query object. This fixes an issue where updating any query param caused all query params to rerun their reactive getters even if they were unchanged.

### DIFF
--- a/src/useRouteQueryParam/formats/RouteParam.ts
+++ b/src/useRouteQueryParam/formats/RouteParam.ts
@@ -58,7 +58,7 @@ export abstract class RouteParam<T> {
   }
 
   public set(query: LocationQuery, value: T | T[] | undefined): void {
-    if (value === undefined) {
+    if (value === undefined || value === null) {
       delete query[this.key]
       return
     }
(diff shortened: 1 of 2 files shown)


src/useRouteQueryParam/formats/RouteParam.ts
```diff
@@ -58,7 +58,7 @@ export abstract class RouteParam<T> {
   }
 
   public set(query: LocationQuery, value: T | T[] | undefined): void {
-    if (value === undefined) {
+    if (value === undefined || value === null) {
```
znicholasbrown, line 61: this isn't strictly part of the fix but we use `Nullable*` for lots of our route classes which meant that they were never getting removed - happy to revert this

pleek91: personally I think we lose functionality here. But maybe this isn't how its working? Hard to remember. 
- 1 - `?foo=1`
- null - `?foo`
- undefined - `?`

znicholasbrown: Hm, maybe it is working like that, that wasn't intuitive to me; I'm gonna revert since it seems like this is intended behavior at the least

pleek91: I believe that reflects how vue-router works as well.

pleek91 (approved):

PR #426, opened 2024-04-09
Enhancement: Boolean route param behavior

znicholasbrown:
This PR updates the `BooleanRouteParam` class as well as the behavior of the `RouteParam` factory to less verbosely handle query parameters based on the presence of a key instead of a parsed string. 

Instead of:
```
path.url?value=true&value_without_default=false
```
we get:
```
path.url?value
```

Callers can use this in the same way as before:

```vue
<script lang="ts" setup>
import { useRouteQueryParam, BooleanRouteParam } from '@prefecthq/vue-compositions'

const value = useRouteQueryParam('value', BooleanRouteParam, true)
const valueWithoutDefault = useRouteQueryParam('value_without_default', BooleanRouteParam)
</script>

```


znicholasbrown:
Note that this is a breaking change due to the removal of the `NullableBooleanRouteParam` class; I've only found one instance where we're using this class internally: https://github.com/PrefectHQ/prefect-ui-library/blob/951057bf7b55b46a8b74e88443c06485130ee4e3/src/compositions/filters.ts#L375, and if we decide to merge this I'll release this as a minor bump and note it in the changelog.

PR #497, opened 2024-10-15
Prevent `useStorage` from immediately saving the default value 

pleek91:
# Description
The `watchEffect` runs immediately which caused the default value to get set into storage. Which means even if the value isn't actually set by the user, its already in storage which can pollute storage (which has a limited size). 

Using a `watch` will only update storage when the value is changed by the user.


znicholasbrown (approved):
Nice

